PR #2, opened 2022-06-21
Styles configuration

tiltom:
Resolves https://sovryn.monday.com/boards/2218344956/pulses/2805419818


apps/frontend/package.json
```diff
@@ -13,19 +14,28 @@
...
+    "mini-css-extract-plugin": "^2.6.1",
+    "postcss": "^8.4.14",
+    "tailwindcss": "^3.1.3",
+    "typescript": "^4.7.4",
+    "postcss-normalize": "^10.0.1",
+    "postcss-import": "^14.1.0",
+    "postcss-nested": "^5.0.6",
+    "postcss-flexbugs-fixes": "^5.0.2",
+    "postcss-preset-env": "^7.7.1",
+    "postcss-nesting": "^10.1.8",
```
creed-victor: I think most of these could be moved to packages/tailwind lib.
And @sovryn/tailwind should be added as dependency here

creed-victor (changes requested):

PR #3, opened 2022-06-23
feat: storybook config

grinry:


README.md
```diff
@@ -1,77 +1,27 @@
...
-- [ESLint](https://eslint.org/) for code linting
-- [Prettier](https://prettier.io) for code formatting
-
-## Setup
-
-This repository is used in the `npx create-turbo` command, and selected when choosing which package manager you wish to use with your monorepo (Yarn).
+- `frontend`: a react app
+- `@sovryn/ui`: a stub React component library shared by our apps
+- `@sovryn/eslint-config-custom`: `eslint` configurations (includes `eslint-config-next` and `eslint-config-prettier`)
+- `@sovryn/tailwindcss-config`: `eslint` configuration
```
tiltom:
```suggestion
- `@sovryn/tailwindcss-config`: `tailwind` configuration
```

packages/ui/.storybook/preview.js
```diff
@@ -0,0 +1,73 @@
...
+
+const screens = {
+  xs: '320px',
+  sm: '576px',
+  md: '768px',
+  lg: '992px',
+  xl: '1200px',
+  '2xl': '1536px',
+  '3xl': '1854px',
+};
```
tiltom:
```suggestion
import '../styles/index.css';
import resolveConfig from 'tailwindcss/resolveConfig';
import tailwindConfig from '@sovryn/tailwindcss-config/index';

const config = resolveConfig(tailwindConfig);

const screens = {
  xs: '320px',
  ...config.theme.screens,
};
```

tiltom: We shouldn't hardcode Tailwind config values but use the actual config values we have defined - see https://tailwindcss.com/docs/configuration#referencing-in-java-script for reference.

packages/ui/.storybook/preview.js
```diff
@@ -0,0 +1,73 @@
...
+export const parameters = {
+  actions: { argTypesRegex: "^on[A-Z].*" },
+  controls: {
+    matchers: {
+      color: /(background|color)$/i,
+      date: /Date$/,
+    },
+  },
+  // values of tailwind
+  backgrounds: {
```
tiltom, line 23: The same as with `screens`, use the actual config values.

packages/ui/src/1_atoms/Button/Button.tsx
```diff
@@ -0,0 +1,9 @@
+import React, { FC } from 'react';
+
+type ButtonProps = {
```
tiltom, line 5: I know this is not very important for the PR but pls, add `primary`, `size` and `label` as properties because in underlines them in the story.

packages/ui/src/Introduction.stories.mdx
```diff
@@ -0,0 +1,211 @@
+import { Meta } from '@storybook/addon-docs';
```
tiltom: Do you want to keep this file? I'd probably deleted it.

tiltom (changes requested):

packages/ui/src/Introduction.stories.mdx
creed-victor: Only if we have something to add there. Currently we have no content for it so I will delete it and we will recover it if we will have somthing

tiltom (approved):

soulBit (approved):

PR #11, opened 2022-09-14
SOV-108: ethers config

creed-victor:
https://sovryn.atlassian.net/browse/SOV-108


apps/frontend/package.json
```diff
@@ -3,11 +3,13 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
+    "@sovryn/ethers-provider": "*",
+    "@sovryn/tailwindcss-config": "*",
+    "@sovryn/ui": "*",
+    "ethers": "^5.7.1",
     "react": "^18.1.0",
     "react-dom": "^18.1.0",
```
tiltom:
```suggestion
    "ethers": "5.7.1",
    "react": "18.1.0",
    "react-dom": "18.1.0",
```

See https://sovryn.monday.com/docs/2733202752?blockId=aebe8f4b-5e4a-4ef6-8967-503ce7a1bb30 for explanation.

apps/frontend/src/app/3_organisms/EthersProviderTest.tsx
```diff
@@ -0,0 +1,66 @@
+import React, { useEffect, useState } from 'react';
+
+import init, { Chain, ChainIds, getProvider } from '@sovryn/ethers-provider';
+
+const chains: Chain[] = [
```
tiltom: I am just wondering, where do we want to have this config? As I understand, `ethers-provider` package is similar to `ui` package - they provide an unopinionated functionality so it seems like we want to have this config in the `frontend` app. If so, I'd suggest moving it to `utils` folder. 

What do you think?

apps/frontend/src/app/3_organisms/EthersProviderTest.tsx
```diff
@@ -0,0 +1,66 @@
...
+    rpcUrl: 'https://public-node.testnet.rsk.co',
+  },
+];
+
+init(chains);
+
+const ProviderItem = ({ chain }: { chain: Chain }) => {
+  const [blockNumber, setBlockNumber] = useState<number>(0);
+  const [date, setDate] = useState<Date>(new Date());
+  const [count, setCount] = useState<number>(0);
```
tiltom, line 8:
```suggestion
  const [blockNumber, setBlockNumber] = useState(0);
  const [date, setDate] = useState(new Date());
  const [count, setCount] = useState(0);
```

packages/ethers-provider/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "typescript": "^4.8.3"
+  },
+  "files": [
+    "dist"
+  ],
+  "dependencies": {
+    "rxjs": "^7.5.6"
+  },
+  "peerDependencies": {
+    "ethers": "^5.7"
```
tiltom:
```suggestion
    "@babel/preset-typescript": "7.18.6",
    "jest": "29.0.3",
    "tsup": "6.2.3",
    "typescript": "4.8.3"
  },
  "files": [
    "dist"
  ],
  "dependencies": {
    "rxjs": "7.5.6"
  },
  "peerDependencies": {
    "ethers": "5.7"
```

packages/ethers-provider/src/provider.ts
```diff
@@ -0,0 +1,45 @@
...
+  if (!ethersProviders[chain.rpcUrl]) {
+    ethersProviders[chain.rpcUrl] = new providers.StaticJsonRpcProvider(
+      chain.providerConnectionInfo && chain.providerConnectionInfo.url
+        ? chain.providerConnectionInfo
+        : chain.rpcUrl,
+    );
+  }
+
+  return ethersProviders[chain.rpcUrl];
+}
```
tiltom:
```suggestion
export function getProvider(
  chainOrChainId?: ChainId | Chain,
): providers.StaticJsonRpcProvider | null {
  let chain: Chain = null;

  if (chainOrChainId === undefined) {
    chain = state.get().chains[0] as Chain;
  }

  if (typeof chainOrChainId === 'string') {
    chain = state.get().chains.find(c => c.id === chainOrChainId) as Chain;
  }

  if (typeof chainOrChainId === 'object') {
    chain = chainOrChainId as Chain;
  }

  if (!chain) {
     return null;
  }

  if (!ethersProviders[chain.rpcUrl]) {
    ethersProviders[chain.rpcUrl] = new providers.StaticJsonRpcProvider(
      chain.providerConnectionInfo && chain.providerConnectionInfo.url
        ? chain.providerConnectionInfo
        : chain.rpcUrl,
    );
  }

  return ethersProviders[chain.rpcUrl];
}
```

What do you think about it? Or do you have some reason for the ignore comments?

packages/ethers-provider/src/store/index.ts
```diff
@@ -0,0 +1,84 @@
...
+  Observable,
+  Subject,
+} from 'rxjs';
+
+import { APP_INITIAL_STATE } from '../constants';
+import { Action, AppState, Chain } from '../types';
+import { ADD_CHAINS, RESET_STORE, UPDATE_CHAIN } from './constants';
+
+const _store = new BehaviorSubject<AppState>(APP_INITIAL_STATE);
+const _stateUpdates = new Subject<AppState>();
```
tiltom:
```suggestion
const store = new BehaviorSubject<AppState>(APP_INITIAL_STATE);
const stateUpdates = new Subject<AppState>();
```

packages/ethers-provider/src/store/index.ts
```diff
@@ -0,0 +1,84 @@
...
+  const state = _store.getValue();
+  _stateUpdates.next(reducer(state, action));
+};
+
+function select(): Observable<AppState>;
+function select<T extends keyof AppState>(stateKey: T): Observable<AppState[T]>;
+function select<T extends keyof AppState>(
+  stateKey?: keyof AppState,
+): Observable<AppState[T]> | Observable<AppState> {
+  if (!stateKey) return _stateUpdates.asObservable();
```
tiltom:
```suggestion
  if (!stateKey) {
     return _stateUpdates.asObservable();
  }
```
I know that not all of the code guidelines we have are applicable to this package but try to use those you can.

tiltom (changes requested):

packages/ethers-provider/src/provider.ts
creed-victor: Would like to prevent using `?.` checks everytime provider method is called, as it will return null only when there is no config added by dev.
Maybe `return null as unknown as providers.StaticJsonRpcProvider` would work?

tiltom: I don't like the current solution but let's wait for @soulBit to see what he thinks. I'd rather prefer the previous one but I think that if a method can return `null`, it should be clear from its signature and not by looking at the implementation.

creed-victor: Then maybe just throw error, because that case isn't supposed to happen (at least while users themselves can't add their own chains and rpcs)

tiltom: That sounds pretty good.

packages/ethers-provider/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "@sovryn/ethers-provider",
+  "version": "0.0.0",
+  "main": "./dist/index.js",
+  "types": "./dist/index.d.ts",
+  "license": "MIT",
+  "author": {
+    "name": "Rytis Grincevicius",
```
soulBit: Replace author details with Sovryn equivalent :)

soulBit (changes requested):

soulBit (approved):

packages/ethers-provider/src/provider.ts
soulBit: victor's new solution with explicit undefined value and checks for this seems ok 👌🏻

PR #15, opened 2022-09-27
SOV-459: feat(dialog): dialog restyling and focus trap

creed-victor:
https://sovryn.atlassian.net/browse/SOV-459


Rickk137:
@creed-victor 
I think the design link that you've implemented the dialog header is not updated.
Take a look at the dialog header style in tx approval dialog:
https://www.figma.com/file/DrjD9LSnpcKyC0OSIEqMD2/Transaction-approval-dialog?node-id=0%3A1

Rickk137:
Dialog edges should be rounded
Screen Shot 2022-09-28 at 2 27 14 PM
Screen Shot 2022-09-28 at 2 27 22 PM

creed-victor:
> @creed-victor I think the design link that you've implemented the dialog header is not updated. Take a look at the dialog header style in tx approval dialog: https://www.figma.com/file/DrjD9LSnpcKyC0OSIEqMD2/Transaction-approval-dialog?node-id=0%3A1

Yeah, I see they now updated figma with improved design - I thought wrapper where component description was written is dialog itself :)

packages/tailwindcss-config/index.js
```diff
@@ -117,6 +117,8 @@ module.exports = {
       'trade-short-25': '#75200D', // '#D74E0940',
       'trade-short-10': '#4F1209', // '#D74E091a',
 
+      'dialog-header': '#292D36',
```
tiltom: I don't like that, have you discussed it with Gilad? We shouldn't add any additional colors to our palette until absolutely necessary and even if we wanted to add it `dialog-header` is so specific that it cannot (and shouldn't) be used anywhere else (so it should be something like `gray-11`, just an example).

Pls discuss it with Gilad.

packages/ui/src/2_molecules/Dialog/Dialog.module.css
```diff
@@ -12,21 +10,12 @@
 }
 
 .dialog {
-  @apply relative bg-black text-sov-white px-4 py-8 mx-auto min-h-full;
+  @apply relative bg-sovryn-black text-sov-white p-0 mx-auto min-h-full;
```
tiltom:
```suggestion
  @apply relative bg-sovryn-black text-gray-10 p-0 mx-auto min-h-full;
```
I think it should be `gray-10` according to the Figma.

packages/ui/src/2_molecules/Dialog/Dialog.tsx
```diff
@@ -28,9 +37,13 @@ export const Dialog: IDialogFunctionComponent<DialogProps> = ({
   dataActionId,
   overlayProps,
   onClose,
+  closeOnEscape = true,
+  initialFocusRef,
 }) => {
   const sizeClassNames = useMemo(() => dialogSizeMap[width], [width]);
```
tiltom: Unnecessary, use it directly.

packages/ui/src/2_molecules/Dialog/Dialog.tsx
```diff
@@ -28,9 +37,13 @@ export const Dialog: IDialogFunctionComponent<DialogProps> = ({
...
 
+  const ref = useRef<HTMLDivElement>(null);
+
   const handleClose = useCallback(() => {
     onClose?.();
   }, [onClose]);
```
tiltom:
```suggestion
  const handleClose = useCallback(() => onClose?.(), [onClose]);
```

packages/ui/src/2_molecules/Dialog/Dialog.tsx
```diff
@@ -28,9 +37,13 @@ export const Dialog: IDialogFunctionComponent<DialogProps> = ({
   dataActionId,
   overlayProps,
   onClose,
+  closeOnEscape = true,
+  initialFocusRef,
```
tiltom, line 35: This `ref` is not used anywhere.

packages/ui/src/2_molecules/Dialog/components/DialogBody.tsx
```diff
@@ -0,0 +1,7 @@
+import { FC, PropsWithChildren } from 'react';
+
+import styles from './DialogBody.module.css';
+
+export const DialogBody: FC<PropsWithChildren> = ({ children }) => {
+  return <main className={styles.container}>{children}</main>;
+};
```
tiltom:
```suggestion
export const DialogBody: FC<PropsWithChildren> = ({ children }) => (
  <main className={styles.container}>{children}</main>
);
```

packages/ui/src/2_molecules/Dialog/components/DialogBody.tsx
```diff
@@ -0,0 +1,7 @@
+import { FC, PropsWithChildren } from 'react';
+
+import styles from './DialogBody.module.css';
+
+export const DialogBody: FC<PropsWithChildren> = ({ children }) => {
```
tiltom: Create a separate folder for `DialogBody` component.

packages/ui/src/2_molecules/Dialog/components/DialogBody.module.css
```diff
@@ -0,0 +1,3 @@
+.container {
```
tiltom: You don't need this file at all, just apply the classes directly. Also, I don't think you need to set text color in dialog body as the only place where it needs to be set is dialog header.

packages/ui/src/2_molecules/Dialog/components/DialogHeader.module.css
```diff
@@ -0,0 +1,11 @@
+.container {
+  @apply px-6 py-3.5 relative bg-dialog-header flex
+   flex-row justify-between items-center space-x-4;
+   @screen sm {
+     @apply rounded-t;
+   }
+}
+
+.title {
+  @apply text-white font-bold text-sm;
```
tiltom:
```suggestion
  @apply text-gray-10 font-bold text-sm;
```
According to the Figma it should be `gray-10`.

packages/ui/src/2_molecules/Dialog/components/DialogHeader.tsx
```diff
@@ -0,0 +1,34 @@
...
+        <button onClick={onClose} data-action-id={dataActionId}>
+          <Icon icon="x-mark" size={12} />
+          <span className="sr-only">Close</span>
+        </button>
+      )}
+    </header>
+  );
+};
+
+export const DefaultTitle: FC<PropsWithChildren> = ({ children }) => (
```
tiltom:
```suggestion 
const DefaultTitle: FC<PropsWithChildren> = ({ children }) => (
```
No need to export this.

packages/ui/src/2_molecules/Dialog/components/DialogHeader.tsx
```diff
@@ -0,0 +1,34 @@
...
+      {renderTitle(title)}
+      {onClose && (
+        <button onClick={onClose} data-action-id={dataActionId}>
+          <Icon icon="x-mark" size={12} />
+          <span className="sr-only">Close</span>
+        </button>
+      )}
+    </header>
+  );
+};
```
tiltom:
```suggestion
export const DialogHeader: FC<DialogHeaderProps> = ({
  title,
  dataActionId,
  onClose,
  renderTitle = title => <DefaultTitle>{title}</DefaultTitle>,
}) => (
  <header className={styles.container}>
    {renderTitle(title)}
    {onClose && (
      <button onClick={onClose} data-action-id={dataActionId}>
        <Icon icon="x-mark" size={12} />
        <span className="sr-only">Close</span>
      </button>
    )}
  </header>
);
```

packages/ui/src/2_molecules/Dialog/components/DialogHeader.module.css
```diff
@@ -0,0 +1,11 @@
+.container {
```
tiltom: You don't need this file, just apply the classes directly.

tiltom (changes requested):

packages/tailwindcss-config/index.js
soulBit: agree with tiltom here, let's add this as generic color to our palette instead

soulBit:
reviewed and no additional comments beyond the ones tiltom already suggested, and the change I made to padding on dialog body - can be sent to QA once changes are resolved and tiltom re-reviews

packages/ui/src/2_molecules/Dialog/Dialog.tsx
creed-victor: Ops, made sure it's used now

soulBit (approved):
All changes completed but component folder refactoring was left out, so I included that and global Roboto definition - PR fully satisfies spec now

packages/tailwindcss-config/index.js
```diff
@@ -96,6 +96,7 @@ module.exports = {
       'gray-8': '#a2a2a2',
       'gray-9': '#c4c4c4',
       // 'gray-10': '#d9d9d9',
+      'gray-11': '#292D36',
```
tiltom: I just said `gray-11` as an example :D The naming itself should come from Gilad but since it's a really dark color, I'd say it will be something like `gray-1.5`.

I am gonna approve the PR for testing but ask Gilad what should be the name and change it later pls.

packages/ui/src/2_molecules/Dialog/Dialog.tsx
```diff
@@ -44,6 +47,35 @@ export const Dialog: IDialogFunctionComponent<DialogProps> = ({
...
+    return () => {
+      document.removeEventListener('keydown', handleEscape);
+    };
+  }, [handleClose, closeOnEscape]);
+
+  useEffect(() => {
+    if (isOpen && initialFocusRef && initialFocusRef.current) {
+      initialFocusRef.current.focus();
+    }
+  }, [isOpen]);
```
tiltom:
```suggestion
  }, [initialFocusRef, isOpen]);
```

tiltom (changes requested):

packages/ui/src/2_molecules/Dialog/Dialog.tsx
creed-victor: Normally  adding `initialFocusRef` as dependency would be unnecessary because it's a ref and is mutable value. 
Do behaviour changes here because it's passed as a prop?

tiltom: I don't know but my linter is screaming at me :D

creed-victor: will make some testing, docs doesn't say anything about refs as prop :)

creed-victor: Ok, including it doesn't break anything so added as dependency

tiltom: @creed-victor btw, if you use`ref` as a `prop`, it acts as a normal property. That's one disadvantage of this approach. If you want, you can use `forwardRef` and it will work as you'd expect (but it's not necessary in this case).

tiltom (approved):

packages/tailwindcss-config/index.js
creed-victor: verified with gilad, gray-90 will be used

PR #17, opened 2022-09-28
SOV-107 - base GTM configuration

soulBit:
https://sovryn.atlassian.net/browse/SOV-107


apps/frontend/.env
```diff
@@ -0,0 +1 @@
+REACT_APP_GOOGLE_ANALYTICS=GTM-PX36D3G
```
creed-victor: Leave value empty here, or rename `.env` to `.env.example`, or move this env to `netlify.toml`.

Current approach will trigger events in development mode and we do not want it, having it in the `netlify.toml` file you will be able to target branches with different gtag id's.

Example: https://github.com/DistributedCollective/Sovryn-frontend/blob/development/netlify.toml#L21

apps/frontend/craco.config.js
```diff
@@ -1,4 +1,14 @@
+const CracoEnvPlugin = require('craco-plugin-env');
```
creed-victor: CRA loads env out of the box so you dont need plugins:

https://create-react-app.dev/docs/adding-custom-environment-variables/#what-other-env-files-can-be-used

https://create-react-app.dev/docs/adding-custom-environment-variables/#referencing-environment-variables-in-the-html

creed-victor (changes requested):

creed-victor (approved):

creed-victor:
@soulBit added condition in the index.html so osamo script would be loaded only build and not development environments. Also hid google analytics if gtag id is not provided in the env

PR #23, opened 2022-10-04
[NavMenuItem] - implement component

pietro-maximoff:
https://sovryn.atlassian.net/browse/SOV-462

### DIFF
--- a/packages/ui/src/2_molecules/NavMenuItem/NavMenuItem.stories.tsx
+++ b/packages/ui/src/2_molecules/NavMenuItem/NavMenuItem.stories.tsx
@@ -9,16 +9,21 @@ export default {
   component: NavMenuItem,
 };
 
-const Template: Story<ComponentProps<typeof NavMenuItem>> = () => (
+const Template: Story<ComponentProps<typeof NavMenuItem>> = args => (
   <>
-    <NavMenuItem children="Zero" count={10} />
-    <NavMenuItem children="Perpetual" className="mx-5" />
-    <NavMenuItem children="999 notifications" count={999} className="mr-5" />
-    <NavMenuItem children="Selected item" count={2} isActive />
+    <NavMenuItem {...args} />
   </>
 );
 
 export const Basic = Template.bind({});
+Basic.args = {
+  children: 'Menu item',
+  count: 100,
+  className: 'm-2',
+  isActive: false,
+  maxNotificationsCount: 99,
+  dataActionId: '',
+};
 
 const InteractiveTemplate: Story<ComponentProps<typeof NavMenuItem>> = args => {
   const [active, handleToggle] = useReducer(state => !state, false);
@@ -30,4 +35,7 @@ Interactive.args = {
   children: 'Menu item',
   count: 1,
   className: 'm-2',
+  isActive: false,
+  maxNotificationsCount: 99,
+  dataActionId: '',
 };
--- a/packages/ui/src/2_molecules/NavMenuItem/NavMenuItem.stories.tsx
+++ b/packages/ui/src/2_molecules/NavMenuItem/NavMenuItem.stories.tsx
@@ -1,6 +1,6 @@
 import { Story } from '@storybook/react';
 
-import React, { ComponentProps, useCallback, useState } from 'react';
+import React, { ComponentProps, useReducer } from 'react';
 
 import { NavMenuItem } from './NavMenuItem';
 
@@ -21,10 +21,8 @@ const Template: Story<ComponentProps<typeof NavMenuItem>> = () => (
 export const Basic = Template.bind({});
 
 const InteractiveTemplate: Story<ComponentProps<typeof NavMenuItem>> = args => {
-  const [active, setActive] = useState(false);
-  const onClick = useCallback(() => setActive(active => !active), []);
-
-  return <NavMenuItem {...args} onClick={onClick} isActive={active} />;
+  const [active, handleToggle] = useReducer(state => !state, false);
+  return <NavMenuItem {...args} onClick={handleToggle} isActive={active} />;
 };
 
 export const Interactive = InteractiveTemplate.bind({});
--- a/packages/ui/src/2_molecules/NavMenuItem/NavMenuItem.stories.tsx
+++ b/packages/ui/src/2_molecules/NavMenuItem/NavMenuItem.stories.tsx
@@ -21,7 +21,7 @@ Basic.args = {
   count: 100,
   className: 'm-2',
   isActive: false,
-  maxNotificationsCount: 99,
+  maxCount: 99,
   dataActionId: '',
 };
 
@@ -36,6 +36,6 @@ Interactive.args = {
   count: 1,
   className: 'm-2',
   isActive: false,
-  maxNotificationsCount: 99,
+  maxCount: 99,
   dataActionId: '',
 };
--- a/packages/ui/src/2_molecules/NavMenuItem/NavMenuItem.stories.tsx
+++ b/packages/ui/src/2_molecules/NavMenuItem/NavMenuItem.stories.tsx
@@ -1,0 +1,35 @@
+import { Story } from '@storybook/react';
+
+import React, { ComponentProps, useCallback, useState } from 'react';
+
+import { NavMenuItem } from './NavMenuItem';
+
+export default {
+  title: 'Molecule/NavMenuItem',
+  component: NavMenuItem,
+};
+
+const Template: Story<ComponentProps<typeof NavMenuItem>> = () => (
+  <>
+    <NavMenuItem children="Zero" count={10} />
+    <NavMenuItem children="Perpetual" className="mx-5" />
+    <NavMenuItem children="999 notifications" count={999} className="mr-5" />
+    <NavMenuItem children="Selected item" count={2} isActive />
+  </>
+);
+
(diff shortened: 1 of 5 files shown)


packages/ui/src/2_molecules/NavMenuItem/NavMenuItem.tsx
```diff
@@ -0,0 +1,50 @@
...
+      onClick={onClick}
+      className={classNames(styles.navMenuItem, className, {
+        [styles.active]: isActive,
+        [styles.count]: formattedCount,
+      })}
+      ref={ref}
+    >
+      {children}
+      {formattedCount && <Badge content={formattedCount} />}
+    </div>
```
creed-victor: I think `button` instead of `div` would make more sense
```suggestion
    <button
      type="button"
      data-action-id={dataActionId}
      onClick={onClick}
      className={classNames(styles.navMenuItem, className, {
        [styles.active]: isActive,
        [styles.count]: formattedCount,
      })}
      ref={ref}
    >
      {children}
      {formattedCount && <Badge content={formattedCount} />}
    </button>
```

packages/ui/src/2_molecules/NavMenuItem/NavMenuItem.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+import React, { useCallback, useState } from 'react';
+
+import { NavMenuItem } from './NavMenuItem';
+
+const TestComponent = () => {
+  const [selected, setSelected] = useState(false);
+  const onClick = useCallback(() => setSelected(selected => !selected), []);
+  return (
+    <NavMenuItem children="Menu item" onClick={onClick} isActive={selected} />
+  );
```
creed-victor:
```suggestion
  const [selected, handleToggle] = useReducer(state => !state, false);
  return (
    <NavMenuItem children="Menu item" onClick={handleToggle} isActive={selected} />
  );
```

packages/ui/src/2_molecules/NavMenuItem/NavMenuItem.stories.tsx
```diff
@@ -0,0 +1,35 @@
...
+);
+
+export const Basic = Template.bind({});
+
+const InteractiveTemplate: Story<ComponentProps<typeof NavMenuItem>> = args => {
+  const [active, setActive] = useState(false);
+  const onClick = useCallback(() => setActive(active => !active), []);
+
+  return <NavMenuItem {...args} onClick={onClick} isActive={active} />;
+};
```
creed-victor:
```suggestion
  const [active, handleToggle] = useReducer(state => !state, false);

  return <NavMenuItem {...args} onClick={handleToggle} isActive={active} />;
};
```

creed-victor (changes requested):

packages/ui/src/2_molecules/NavMenuItem/NavMenuItem.tsx
```diff
@@ -0,0 +1,50 @@
...
+  children,
+  isActive,
+  className,
+  dataActionId,
+}) => {
+  const ref = useRef<HTMLDivElement>(null);
+  const formattedCount = useMemo(
+    () =>
+      count && count > MAX_NOTIFICATIONS_COUNT
+        ? MAX_NOTIFICATIONS_COUNT
```
creed-victor: if it's more than `99`, it should return `99+`

```suggestion
        ? `${MAX_NOTIFICATIONS_COUNT}+`
```

packages/ui/src/2_molecules/NavMenuItem/NavMenuItem.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+      <NavMenuItem children="Menu item" count={5} />,
+    );
+    expect(getByText('5')).toBeInTheDocument();
+  });
+
+  it('should show the maximum number which is 99 inside the Badge', () => {
+    const { getByText } = render(
+      <NavMenuItem children="Menu item" count={1000} />,
+    );
+    expect(getByText('99')).toBeInTheDocument();
```
creed-victor:
```suggestion
    expect(getByText('99+')).toBeInTheDocument();
```

packages/ui/src/2_molecules/NavMenuItem/NavMenuItem.tsx
```diff
@@ -0,0 +1,51 @@
+import React, { FC, ReactNode, useMemo, useRef } from 'react';
+
+import classNames from 'classnames';
+
+import { Badge } from '../../1_atoms/Badge/Badge';
+import styles from './NavMenuItem.module.css';
+
+const MAX_NOTIFICATIONS_COUNT = 99;
```
tiltom:
```suggestion
const MAX_COUNT_DEFAULT = 99;
```
I don't like this because you can basically never override it, unless you change it in the UI package and redeploy it. I'd add a property `maxNotificationsCount` that will be optional and its default value would be MAX_COUNT_DEFAULT.

packages/ui/src/2_molecules/NavMenuItem/NavMenuItem.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+export const NavMenuItem: FC<NavMenuItemProps> = ({
+  count,
+  onClick,
+  children,
+  isActive,
+  className,
+  dataActionId,
+}) => {
+  const ref = useRef<HTMLButtonElement>(null);
```
tiltom:
```suggestion

```
This `ref` is useless as you do nothing with it, you just assign it to the button. What you should do is to use `forwardRef` and expose `ref` this way but it will be done in bulk in the upcoming sprint so you don't need to add it now.

tiltom (changes requested):

tiltom:
The component also lacks one very important thing - disabled state. I know it's not designed in Figma and I don't want to block this story but pls create a follow-up story to the next sprint (with 1 SP estimate) and implement it there (after a discussion with Gilad).

packages/ui/src/2_molecules/NavMenuItem/NavMenuItem.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const formattedCount = useMemo(
+    () =>
+      count && count > MAX_NOTIFICATIONS_COUNT
+        ? MAX_NOTIFICATIONS_COUNT
+        : count,
+    [count],
+  );
```
tiltom:
```suggestion
  const formattedCount = useMemo(
    () =>
      count && count > MAX_NOTIFICATIONS_COUNT
        ? `${MAX_NOTIFICATIONS_COUNT}+`
        : count,
    [count],
  );
```

tiltom (changes requested):

packages/ui/src/2_molecules/NavMenuItem/NavMenuItem.tsx
pietro-maximoff: it's better to leave it without `+` because we use fixed `width` and `height` in the Badge component, or I need to adjust the Badge component according to this

pietro-maximoff:
> The component also lacks one very important thing - a disabled state. I know it's not designed in Figma and I don't want to block this story but pls create a follow-up story to the next sprint (with 1 SP estimate) and implement it there (after a discussion with Gilad).

@tiltom why do we need a disabled state here? i thought we will have maintenance mode for all pages, otherwise, we need to add  some tooltip with an explanation why this button is disabled because it's unclear especially if you have a notification count there

packages/ui/src/2_molecules/NavMenuItem/NavMenuItem.tsx
tiltom: But it won't work as intended without `+`. It's written in the spec and it also doesn't make logical sense to display let's say `99` if you have `1000` notifications. If there are changes needed in the Badge component, do them pls.

tiltom:
> > The component also lacks one very important thing - a disabled state. I know it's not designed in Figma and I don't want to block this story but pls create a follow-up story to the next sprint (with 1 SP estimate) and implement it there (after a discussion with Gilad).
> 
> @tiltom why do we need a disabled state here? i thought we will have maintenance mode for all pages, otherwise, we need to add some tooltip with an explanation why this button is disabled because it's unclear especially if you have a notification count there

We used it for instance with Perps - prior to launch we had it disabled in the navigation and then we linked a page to it. You can wait for Gilad and discuss it with him if it's needed, just take a note and bring it up once he's back.

packages/ui/src/2_molecules/NavMenuItem/NavMenuItem.tsx
```diff
@@ -0,0 +1,51 @@
...
+const MAX_COUNT_DEFAULT = 99;
+
+type NavMenuItemProps = {
+  count?: number;
+  onClick?: () => void;
+  children: ReactNode;
+  isActive?: boolean;
+  className?: string;
+  dataActionId?: string;
+  maxNotificationsCount?: number;
```
tiltom:
```suggestion
  maxCount?: number;
```

tiltom (changes requested):

tiltom (approved):

creed-victor (approved):

LunaSov (approved):

PR #25, opened 2022-10-06
chore: repo config

creed-victor:
https://sovryn.atlassian.net/browse/SOV-675

- ui package is bundled
- husky prettifies all staged files before commit
- eslint and jest checks are executed on staged files before commit
- changed components to use css module classes instead of tailwind directly


soulBit (approved):

packages/ui/src/2_molecules/Overlay/Overlay.module.css
```diff
@@ -0,0 +1,47 @@
...
+.overlay {
+  @apply inset-0 flex flex-col;
+
+  &.__fixed {
+    @apply fixed;
+  }
+
+  &.__absolute {
+    @apply absolute;
+  }
```
tiltom, line 14:
```suggestion
  &.fixed {
    @apply fixed;
  }

  &.absolute {
    @apply absolute;
  }
```
No need to follow `BEM` here.

tiltom (changes requested):

packages/ui/src/2_molecules/Overlay/Overlay.module.css
creed-victor: `fixed` and `absolute` are existing default classes on tailwind - will cause error, happy to change to some different name

tiltom: Ah okay, I see that now. In that case it's fine and you can leave it as is :)

tiltom (approved):

PR #38, opened 2022-10-18
Sov 144 d 2 transaction id implementation

Rickk137:
https://sovryn.atlassian.net/browse/SOV-144


packages/ui/src/2_molecules/TransactionID/TransactionID.tsx
```diff
@@ -0,0 +1,89 @@
...
+interface TransactionIDProps {
+  value: string;
+  startLength?: number;
+  endLength?: number;
+  className?: string;
+  hideTooltip?: boolean;
+  chainId?: number | ChainId;
+  dataLayoutId?: string;
+  isNativeBtc?: boolean;
+  isAddress?: boolean;
```
tiltom, line 21: This prop is not listed in the story among other props. If it should be there, pls adjust it.

If it should stay, I'd probably adjust it because its purpose is not clear at a first sight -  I suggest enum (e.g. TransactionIdUsage) with values `address` and `transaction`. It's also more scalable if we decide to add more use cases in the future.

What do you think @soulBit ?

packages/ui/src/2_molecules/TransactionID/TransactionID.tsx
```diff
@@ -0,0 +1,89 @@
...
+import classNames from 'classnames';
+
+import { Icon } from '../../1_atoms';
+import { COPY, NEW_TAB } from '../../1_atoms/Icon/iconNames';
+import { Tooltip } from '../Tooltip/Tooltip';
+import { TooltipPlacement, TooltipTrigger } from '../Tooltip/Tooltip.types';
+import styles from './TransactionID.module.css';
+import { blockExplorers, ChainId, currentChainId } from './TransactionID.types';
+
+interface TransactionIDProps {
```
tiltom:
```suggestion
type TransactionIdProps {
```

packages/ui/src/2_molecules/TransactionID/TransactionID.tsx
```diff
@@ -0,0 +1,89 @@
...
+  endLength?: number;
+  className?: string;
+  hideTooltip?: boolean;
+  chainId?: number | ChainId;
+  dataLayoutId?: string;
+  isNativeBtc?: boolean;
+  isAddress?: boolean;
+}
+
+export const TransactionID: React.FC<TransactionIDProps> = ({
```
tiltom:
```suggestion
export const TransactionId: React.FC<TransactionIdProps> = ({
```

packages/ui/src/2_molecules/TransactionID/TransactionID.tsx
```diff
@@ -0,0 +1,89 @@
...
+  endLength = 4,
+  className,
+  hideTooltip,
+  chainId,
+  dataLayoutId,
+  isNativeBtc,
+  isAddress,
+  value,
+}) => {
+  const transactionId = useMemo(() => {
```
tiltom:
```suggestion
  const formattedValue = useMemo(() => {
```

packages/ui/src/2_molecules/TransactionID/TransactionID.tsx
```diff
@@ -0,0 +1,89 @@
...
+    if (value?.length && startLength && endLength) {
+      const start = value.substr(0, startLength);
+      const end = value.substr(-endLength);
+      return `${start} ··· ${end}`;
+    }
+    return value;
+  }, [value, startLength, endLength]);
+
+  const url = useMemo(() => {
```
tiltom:
```suggestion
  const blockExplorerUrl = useMemo(() => {
```

packages/ui/src/2_molecules/TransactionID/TransactionID.tsx
```diff
@@ -0,0 +1,89 @@
...
+    if (isNativeBtc) {
+      return blockExplorers[`btc_${currentChainId}`];
+    }
+    return blockExplorers[chainId || currentChainId];
```
tiltom, line 48: This is really misleading - you're implying there is some detection of `currentChainId` but under the hood, you're just hardcoding it to `30`. What you should do instead is to create a local const `DEFAULT_CHAIN_ID`, set it to `ChainId.RSK_MAINNET` and delete `currentChainId`.

packages/ui/src/2_molecules/TransactionID/TransactionID.tsx
```diff
@@ -0,0 +1,89 @@
...
+  }, [isNativeBtc, chainId]);
+
+  const copyAddress = useCallback(async () => {
+    await navigator.clipboard.writeText(value);
+    alert('Address was copied to clipboard.');
+  }, [value]);
```
tiltom, line 54:
```suggestion
  const onCopyAddress = useCallback(
    () => navigator.clipboard.writeText(value),
    [value],
  );
```
I am not sure if we want to have `alert` after copy or not, pls discuss it with Gilad.

packages/ui/src/2_molecules/TransactionID/TransactionID.tsx
```diff
@@ -0,0 +1,89 @@
...
+        <span className="flex items-center">
+          {value}
+          <span
+            data-layout-id="transaction-copy"
+            className={styles.icon}
+            onClick={copyAddress}
+          >
+            <Icon icon={COPY} />
+          </span>
+          <a
```
tiltom, line 68: Try if you can use our `Link` component. This is the exact use case why we have it. You may have some problems with its default styling but maybe you'll be able to override it.

packages/ui/src/2_molecules/TransactionID/TransactionID.types.ts
```diff
@@ -0,0 +1,29 @@
+export const chains = {
+  mainnet: 30,
+  testnet: 31,
+};
+
+export const currentChainId = chains.mainnet;
```
tiltom, line 6:
```suggestion
```
You don't need those, they had some purpose in our dapp v1 but they are unnecessary in the UI package, `ChainId` is all you need.

packages/ui/src/2_molecules/TransactionID/TransactionID.module.css
```diff
@@ -0,0 +1,15 @@
+.link {
```
tiltom, line 1: Link should be without opacity if selected.

Implementation:
image

Design:
image

tiltom (changes requested):

packages/ui/src/2_molecules/TransactionID/TransactionID.tsx
soulBit: this can be left in for now, we'll be added a Toast component soon to handle notifications like this

PR #44, opened 2022-10-20
SOV-678: add address table component

tiltom:
Resolves https://sovryn.atlassian.net/browse/SOV-678

### DIFF
--- a/packages/ui/src/2_molecules/TableBase/TableBase.stories.tsx
+++ b/packages/ui/src/2_molecules/TableBase/TableBase.stories.tsx
@@ -10,18 +10,18 @@ const columns = [
   {
     id: 'index',
     title: 'Index',
-    align: Align.center,
+    align: Align.left,
     cellRenderer: row => `${row.index}.`,
   },
   {
     id: 'address',
     title: 'Address',
-    align: Align.center,
+    align: Align.left,
   },
   {
     id: 'balance',
     title: 'Balance',
-    align: Align.center,
+    align: Align.left,
     cellRenderer: row => `${row.balance} RBTC`,
   },
 ];
@@ -46,7 +46,7 @@ export default {
 };
 
 const Template: Story<ComponentProps<typeof TableBase>> = args => (
-  <div className="max-w-lg">
+  <div className="max-w-sm">
     <TableBase {...args} />
   </div>
 );
--- a/packages/ui/src/2_molecules/TableBase/TableBase.stories.tsx
+++ b/packages/ui/src/2_molecules/TableBase/TableBase.stories.tsx
@@ -1,0 +1,71 @@
+import { Story } from '@storybook/react';
+
+import React, { ComponentProps } from 'react';
+
+import { prettyTx } from '../../utils';
+import { TableBase } from './TableBase';
+import { Align } from './TableBase.types';
+
+const columns = [
+  {
+    id: 'index',
+    title: 'Index',
+    align: Align.center,
+    cellRenderer: row => `${row.index}.`,
+  },
+  {
+    id: 'address',
+    title: 'Address',
+    align: Align.center,
+  },
+  {
+    id: 'balance',
+    title: 'Balance',
+    align: Align.center,
+    cellRenderer: row => `${row.balance} RBTC`,
+  },
+];
+
+// TODO: Change hardcoded addresses for TransactionId component once it's merged
+const rows = [
+  {
+    index: 1,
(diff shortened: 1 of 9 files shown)


packages/ui/src/2_molecules/TableBase/TableBase.stories.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+import { prettyTx } from '../../utils';
+import { TableBase } from './TableBase';
+import { Align } from './TableBase.types';
+
+const columns = [
+  {
+    id: 'index',
+    title: 'Index',
+    align: Align.center,
```
pietro-maximoff: pls replace all columns with `Align.left`, it should be by default, and need adjustments with the table header, when we change it to the left position, see pics:

Current version:
worktable
Design version:
tableOriginal

packages/ui/src/2_molecules/TableBase/components/TableRow/TableRow.module.css
```diff
@@ -0,0 +1,45 @@
+.row {
+  @apply cursor-pointer sm:cursor-auto outline outline-1 outline-gray-70 rounded bg-gray-90 text-xs font-medium;
+
+  & > td {
+    @apply py-3 px-4;
```
pietro-maximoff:
```suggestion
    @apply py-2.5 px-4;
```

pietro-maximoff (changes requested):
found a couple of small issues, but other stuff looks good 👍

packages/ui/src/2_molecules/TableBase/TableBase.tsx
```diff
@@ -0,0 +1,76 @@
...
+import rowStyles from './components/TableRow/TableRow.module.css';
+
+export type TableBaseProps<RowType extends RowObject> = {
+  className?: string;
+  columns: ColumnOptions<RowType>[];
+  rows?: RowType[];
+  rowKey?: (row: RowType) => number | string;
+  noData?: ReactNode;
+  onRowClick?: (row: RowType) => void;
+  dataAttribute?: string;
```
Rickk137, line 17: any reason you've changed ``dataLayoutId`` to this?

Rickk137 (changes requested):

Rickk137:
@tiltom as pietro mentioned i think the headers padding should be similar to the rows. when you want to align items to left they would be aligned properly

Rickk137:
I'm not sure if the design is indicating that when we have no data we must show "-"? wdyt @tiltom 
Screen Shot 2022-10-21 at 12 08 18 PM

packages/ui/src/2_molecules/TableBase/components/TableRow/TableRow.tsx
```diff
@@ -0,0 +1,59 @@
...
+              column.align && [styles[`text${column.align}`]],
+            )}
+          >
+            {column.cellRenderer
+              ? column.cellRenderer(row, column.id)
+              : row[column.id]}
+          </td>
+        ))}
+      </tr>
+      <tr className={styles.spacer}></tr>
```
Rickk137, line 56: wouldn't it be simpler to give bottom padding to the ``table tr -> th`` ?

Rickk137 (changes requested):

packages/ui/src/2_molecules/TableBase/TableBase.tsx
tiltom: yes, because we discussed it in https://sovryn.atlassian.net/browse/SOV-861 so I wanted to save soulBit some time :D

tiltom:
> I'm not sure if the design is indicating that when we have no data we must show "-"? wdyt @tiltom Screen Shot 2022-10-21 at 12 08 18 PM

No, what you are referring to is a state when you don't have all of the properties for one row but in that case, you can manage it either in `cellRenderer` or in the object itself. If you have no data (= no rows), there will be `noData` prop used as you wouldn't have any indices whatsoever.

packages/ui/src/2_molecules/TableBase/components/TableRow/TableRow.tsx
tiltom: Yeah, I wish it was that simple :D But this is a table - `padding` is applied only in the cell itself and `margin` is ignored. I don't like this solution either but I could choose between more hacky and less hacky options.

Rickk137 (approved):

pietro-maximoff (approved):

LunaSov (approved):

PR #45, opened 2022-10-20
Feat/sov 789 d 2 header component

pietro-maximoff:
https://sovryn.atlassian.net/browse/SOV-789


packages/ui/src/1_atoms/Button/Button.test.tsx
```diff
@@ -65,15 +66,14 @@ describe('Button', () => {
...
-  //   expect(button).toContain('large');
-  // });
-  // it('should render a button style with a className equal to the secondary', () => {
-  //   render(<Button text="Button style" style={ButtonStyle.secondary} />);
-  //   const classes = screen.getByText('Button style').getAttribute('class');
-  //   expect(button).toContain('secondary');
-  // });
+  it('should render a button size with a className equal to the large', () => {
+    render(<Button text="Button size" size={ButtonSize.large} />);
+    const classes = screen.getByText('Button size').getAttribute('class');
```
tiltom:
```suggestion
    const {getByText} = render(<Button text="Button size" size={ButtonSize.large} />);
    const classes = getByText('Button size').getAttribute('class');
```

packages/ui/src/1_atoms/Button/Button.test.tsx
```diff
@@ -65,15 +66,14 @@ describe('Button', () => {
...
+    expect(classes).toContain('large');
+  });
+  it('should render a button style with a className equal to the secondary', () => {
+    render(<Button text="Button style" style={ButtonStyle.secondary} />);
+    const classes = screen.getByText('Button style').getAttribute('class');
```
tiltom:
```suggestion
    const {getByText} = render(<Button text="Button style" style={ButtonStyle.secondary} />);
    const classes = getByText('Button style').getAttribute('class');
```

packages/ui/src/2_molecules/Header/Header.module.css
```diff
@@ -0,0 +1,19 @@
+.header {
+  @apply bg-gray-90 h-16 w-full flex align-middle justify-between px-11;
```
tiltom, line 2:
```suggestion
  @apply bg-gray-90 h-16 w-full flex align-middle justify-between px-12;
```

packages/ui/src/2_molecules/Header/Header.module.css
```diff
@@ -0,0 +1,19 @@
...
+
+  & > div {
+    @apply flex items-center;
+  }
+}
+
+.logo {
```
tiltom, line 9: This isn't used anywhere but I guess it should be because logo has some padding in the design (and if not, apply the suggestions in this file so we have the desired spacing).

packages/ui/src/2_molecules/Header/Header.module.css
```diff
@@ -0,0 +1,19 @@
...
+  @apply max-h-7;
+}
+
+.menuItems {
+  @apply mx-11 flex items-center;
```
tiltom:
```suggestion
  @apply mx-12 flex items-center;
```

packages/ui/src/2_molecules/Header/Header.tsx
```diff
@@ -0,0 +1,27 @@
...
+  return (
+    <header data-layout-id={dataLayoutId} className={styles.header}>
+      <div>
+        {logo}
+        {menuItems && <div className={styles.menuItems}>{menuItems}</div>}
+      </div>
+      {secondaryContent && <div>{secondaryContent}</div>}
+    </header>
+  );
+};
```
tiltom:
```suggestion
}) => (
  <header data-layout-id={dataLayoutId} className={styles.header}>
    <div>
      {logo}
      {menuItems && <div className={styles.menuItems}>{menuItems}</div>}
    </div>
    {secondaryContent && <div>{secondaryContent}</div>}
  </header>
);
```

packages/ui/src/2_molecules/Header/Header.tsx
```diff
@@ -0,0 +1,27 @@
...
+export const Header: FC<HeaderProps> = ({
+  dataLayoutId,
+  logo,
+  menuItems,
+  secondaryContent,
+}) => {
+  return (
+    <header data-layout-id={dataLayoutId} className={styles.header}>
+      <div>
+        {logo}
```
tiltom: Logo should be clickable but since it's `ReactNode`, we can just wrap the image with `Link` component or `a` tag. Or would you prefer to have some property for it?

tiltom (changes requested):

packages/ui/src/2_molecules/Header/Header.module.css
pietro-maximoff: its related to my header comment

packages/ui/src/2_molecules/Header/Header.tsx
pietro-maximoff: nope, will add it

packages/ui/src/2_molecules/Header/Header.module.css
pietro-maximoff: It just need to be deleted, padding added in the header and in menuItems

packages/ui/src/2_molecules/Header/Header.module.css
pietro-maximoff: @tiltom since we have a margin in the logo with 44px from both sides, I believe it would be more sensible to leave it 44px from the right side of the header

packages/ui/src/2_molecules/Header/Header.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export const Header: FC<HeaderProps> = ({
+  dataLayoutId,
+  logo,
+  menuItems,
+  secondaryContent,
+}) => (
+  <header data-layout-id={dataLayoutId} className={styles.header}>
+    <div>
+      {logo && <Link className={styles.logo} href="/" text={logo} />}
```
tiltom: This is not good, I'd probably just pass the whole `Link` with `img` in `logo` property because you can customize the link and also you probably don't want to use `href` for navigating in the dapp (and we just got rid of `react-router` in the UI package).

What do you think @soulBit ?

tiltom (changes requested):

packages/ui/src/2_molecules/Header/Header.tsx
pietro-maximoff: Yep, agree with that

tiltom (approved):

soulBit (approved):

LunaSov (approved):

PR #60, opened 2022-10-31
SOV-813: contract config setup

creed-victor:
https://sovryn.atlassian.net/browse/SOV-813


apps/frontend/src/config/networks.ts
```diff
@@ -0,0 +1,10 @@
+import { ChainIds } from '@sovryn/ethers-provider';
+
+export const networkMap = {
+  [ChainIds.RSK_MAINNET]: 'rsk',
+  [ChainIds.RSK_TESTNET]: 'rskTestnet',
+  [ChainIds.MAINNET]: 'eth',
+  [ChainIds.ROPSTEN]: 'ropsten',
```
soulBit, line 7: We should use Sepolia for Eth testnet going forward

apps/frontend/src/config/networks.ts
```diff
@@ -0,0 +1,10 @@
+import { ChainIds } from '@sovryn/ethers-provider';
+
+export const networkMap = {
```
soulBit, line 3: Do we want to include a map of default gas settings for each chain as well?

soulBit (changes requested):

Rickk137:
@soulBit @creed-victor We can move the sovryn-contracts to a new package.

apps/frontend/src/config/networks.ts
tiltom: I think that's a good idea @soulBit :)

apps/frontend/src/config/contracts/index.ts
```diff
@@ -0,0 +1,11 @@
...
+import loanTokens from './loan-tokens';
+import protocol from './protocol';
+
+const contracts = {
+  tokens,
+  loanTokens,
+  protocol,
+} as const;
+
+export default contracts;
```
tiltom: Why do you use `default` export instead of named export?

apps/frontend/src/config/contracts/protocol/index.ts
```diff
@@ -0,0 +1,8 @@
+import { AsyncContractConfigData } from '../../../types/config';
+import rsk from './rsk';
+
+const protocol: Record<string, Record<string, AsyncContractConfigData>> = {
```
tiltom:
```suggestion
const protocol: Partial<
  Record<Network, Record<string, AsyncContractConfigData>>
> = {
```

apps/frontend/src/utils/contracts/global.test.ts
```diff
@@ -0,0 +1,88 @@
...
+      } catch (e) {
+        expect(e.message).toBe(
+          'findContract: Unknown contract: 0x0000000000000000000000000000000000000000',
+        );
+      }
+    });
+  });
+
+  describe('getContract', () => {
+    it('get xusd contract', async () => {
```
tiltom:
```suggestion
    it('gets xusd contract', async () => {
```

tiltom (changes requested):

apps/frontend/src/config/networks.ts
creed-victor: You mean gas prices or gas limit for each transaction?
Gas prices should be always retrieved from the block before transaction is being submitted and we will be able to estimate gas limits here because new tx flow asks first tx to be confirmed before second is submitted

tiltom: I think soulBit meant gas limit, it doesn't make sense to store gas prices.

creed-victor: Got it, in any case I think we don't need it here at this point.
D2 will be used for zero only at first and we will be waiting for transactions to be confirmed before sending another one (at least that's how I understand design), so it solves issue we had in D1

@tiltom @soulBit

apps/frontend/src/config/contracts/index.ts
creed-victor: A habit to have configs exported as default, will change it

apps/frontend/src/config/networks.ts
creed-victor: added sepolia, but leaving ropsten with deprecation flag for backwards compability

apps/frontend/src/config/contracts/protocol/index.ts
```diff
@@ -0,0 +1,11 @@
+import { AsyncContractConfigData } from '../../../types/config';
+import { Network } from '../../networks';
+import rsk from './rsk';
+
+const protocol: Partial<
```
tiltom:
```suggestion
export const protocol: Partial<
```

apps/frontend/src/config/contracts/loan-tokens/rskTestnet.ts
```diff
@@ -0,0 +1,5 @@
+const rskTestnet: Record<string, string> = {
```
tiltom:
```suggestion
export const rskTestnet: Record<string, string> = {
```

apps/frontend/src/config/contracts/loan-tokens/rsk.ts
```diff
@@ -0,0 +1,5 @@
+const rsk: Record<string, string> = {
```
tiltom:
```suggestion
export const rsk: Record<string, string> = {
```

apps/frontend/src/config/contracts/loan-tokens/index.ts
```diff
@@ -0,0 +1,10 @@
+import { Network } from '../../networks';
+import rsk from './rsk';
+import rskTestnet from './rskTestnet';
+
+const tokens: Partial<Record<Network, Record<string, string>>> = {
```
tiltom:
```suggestion
export const tokens: Partial<Record<Network, Record<string, string>>> = {
```

apps/frontend/src/config/contracts/protocol/rsk.ts
```diff
@@ -0,0 +1,14 @@
+import type { AsyncContractConfigData } from '../../../types/config';
+
+const rsk: Record<string, AsyncContractConfigData> = {
```
tiltom:
```suggestion
export const rsk: Record<string, AsyncContractConfigData> = {
```

apps/frontend/src/config/contracts/tokens/index.ts
```diff
@@ -0,0 +1,10 @@
+import { Network } from '../../networks';
+import rsk from './rsk';
+import rskTestnet from './rskTestnet';
+
+const tokens: Partial<Record<Network, Record<string, string>>> = {
+  rsk,
+  rskTestnet,
+};
+
+export default tokens;
```
tiltom: Do not use default export anywhere if not necessary pls.

tiltom (changes requested):

Rickk137 (approved):

tiltom (approved):

PR #64, opened 2022-11-03
chore: use data attribute util in new components

soulBit:
Use new utility method for applying data-layout-id attributes to components

### DIFF
--- a/.changeset/curvy-waves-burn.md
+++ b/.changeset/curvy-waves-burn.md
@@ -1,0 +1,5 @@
+---
+"@sovryn/ui": patch
+---
+
+chore: use data attribute util in new components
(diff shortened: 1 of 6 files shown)


Rickk137 (approved):

tiltom (approved):
The code is fine but we discussed in the past that we'd like to use `dataAttribute` instead of `dataLayoutId` in `props`. Do we want to do it in a separate PR or will you add it to this one?

pietro-maximoff (approved):

soulBit:
> The code is fine but we discussed in the past that we'd like to use `dataAttribute` instead of `dataLayoutId` in `props`. Do we want to do it in a separate PR or will you add it to this one?

I'll create a new story for this change, there's quite a few instances

PR #66, opened 2022-11-03
SOV-892: Tab component styling updates

soulBit:
https://sovryn.atlassian.net/browse/SOV-892

### DIFF
--- a/packages/ui/src/2_molecules/Tabs/components/Tab/Tab.module.css
+++ b/packages/ui/src/2_molecules/Tabs/components/Tab/Tab.module.css
@@ -36,13 +36,15 @@
   }
 
   &.secondary {
+    @apply bg-gray-80 mx-1.5;
+
     &.normal {
-      @apply font-semibold text-xs px-0.5 py-2 mx-0.5;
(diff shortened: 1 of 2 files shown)


packages/ui/src/2_molecules/Tabs/components/Tab/Tab.module.css
```diff
@@ -36,13 +36,15 @@
   }
 
   &.secondary {
+    @apply bg-gray-80 mx-1.5;
```
Rickk137, line 39: I think the first and last child should have 0 margin. we can use gap on parent or remove margin for last-first child here.

Rickk137 (changes requested):

packages/ui/src/2_molecules/Tabs/components/Tab/Tab.module.css
soulBit: nice catch, updated

Rickk137 (approved):

pietro-maximoff (approved):

LunaSov (approved):

PR #67, opened 2022-11-04
SOV-828: social links component

soulBit:
https://sovryn.atlassian.net/browse/SOV-828


apps/frontend/src/app/2_molecules/SocialLinks/SocialLinks.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { applyDataAttr, Icon } from '@sovryn/ui';
+
+import { translations } from '../../../locales/i18n';
+import { sovrynLinks } from '../../../utils/constants';
+import styles from './SocialLinks.module.css';
+
+export type SocialLinksProps = {
+  className?: string;
+  innerClassName?: string;
+  dataAttributeId?: string;
```
tiltom:
```suggestion
  dataAttribute?: string;
```

apps/frontend/src/app/2_molecules/SocialLinks/SocialLinks.tsx
```diff
@@ -0,0 +1,70 @@
...
+}) => {
+  const { t } = useTranslation();
+
+  return (
+    <div
+      className={classNames(styles.socials, className)}
+      {...applyDataAttr(dataAttributeId)}
+    >
+      {socials.map(item => (
+        <a
```
tiltom: I know why you did it but it's a bit weird to use `a` tag if we have `Link` component. I am thinking if we don't need something like `InputBase` that would just provide unstyled link functionality so we could use it with everything and then have `Link` that would just add its styling. Let me know if you think it's an overkill :D

tiltom (changes requested):

apps/frontend/src/app/2_molecules/SocialLinks/SocialLinks.tsx
soulBit: yep I was on the fence about this as well, suggestion taken and implemented :)
(it cleans up responsibilities nicely)

tiltom (approved):

apps/frontend/src/app/2_molecules/SocialLinks/SocialLinks.module.css
```diff
@@ -0,0 +1,20 @@
...
+    }
+
+    &:last-child {
+      @apply mr-0;
+    }
+  }
+
+  .ring {
+    @apply border rounded-full p-1.5 flex justify-center items-center hover:bg-gray-80;
+    border-color: rgba(255, 255, 255, 0.1);
```
Rickk137: border-white/10

Rickk137 (approved):

LunaSov (approved):

PR #71, opened 2022-11-08
[VerticalTabsMobile] - add a component

pietro-maximoff:
https://sovryn.atlassian.net/browse/SOV-802

### DIFF
--- a/packages/ui/src/2_molecules/VerticalTabsMobile/VerticalTabsMobile.module.css
+++ b/packages/ui/src/2_molecules/VerticalTabsMobile/VerticalTabsMobile.module.css
@@ -1,0 +1,26 @@
+.container {
+  @apply w-full relative;
+}
+
+.aside {
+  @apply bg-gray-90 py-6 px-7 relative flex flex-col justify-between items-start;
+}
+
+.header {
+  @apply pb-6 w-full relative;
+}
+
+.tabs {
+  @apply w-full flex flex-col justify-center items-center gap-y-3;
+}
+
+.content {
+  @apply w-full h-full bg-gray-90 py-6 px-7 relative overflow-y-auto;
+}
+
+.buttonBack {
+  @apply flex items-center text-sm mb-11;
+  & svg {
+    @apply bg-gray-70 w-6 h-6 p-1 mr-2 rounded flex items-center justify-center;
+  }
+}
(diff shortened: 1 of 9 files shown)


packages/ui/src/2_molecules/VerticalTabsMobile/VerticalTabsMobile.tsx
```diff
@@ -0,0 +1,72 @@
...
+          </div>
+        </aside>
+      ) : (
+        <div
+          className={classNames(styles.content, props.contentClassName)}
+          ref={contentRef}
+        >
+          <button onClick={handleBack} className={styles.buttonBack}>
+            <Icon icon="arrow-back" size={14} /> Back to wallet menu
+          </button>
```
creed-victor: Text assumes that component will be used for wallets only.

I think entire `button` should be removed from component and it should be added together with `renderContent`, so dev could make sure we can use custom back buttons or even prevent them from appearing.

creed-victor (changes requested):

packages/ui/src/2_molecules/VerticalTabsMobile/VerticalTabsMobile.module.css
```diff
@@ -0,0 +1,27 @@
...
+.content {
+  @apply w-full h-full bg-gray-90 py-6 px-7 relative overflow-y-auto;
+  & button {
+    &:first-child {
+      @apply flex items-center text-sm mb-11;
+      & svg {
+        @apply bg-gray-70 w-6 h-6 p-1 mr-2 rounded flex items-center justify-center;
+      }
+    }
+  }
```
creed-victor: Remove it from here and apply classes inline - as it's only used in storybook it's ok on using tailwind classes directly there.

packages/ui/src/2_molecules/VerticalTabsMobile/VerticalTabsMobile.stories.tsx
```diff
@@ -0,0 +1,147 @@
...
+          header={() => <p>Select the type of wallet you have</p>}
+          items={[
+            {
+              label: 'Tab 1',
+              content: (
+                <>
+                  <button onClick={() => setSelectedIndex(null)}>
+                    <Icon icon="arrow-back" size={14} />
+                    Back to wallet menu
+                  </button>
```
creed-victor: It's being repeated multiple times, make it a component and reuse it.

creed-victor (changes requested):

creed-victor (approved):

soulBit (approved):
Approving but we should replace the `<p>` and `<small>` tags in MobileVerticalTabItem component with `<Heading>`, once Victor has made adjustments to CSS overridesfor font color (or in SOV-1027, whichever comes first)

LunaSov (approved):

PR #73, opened 2022-11-08
SOV-931: use Paragraph component for vertical tabs

creed-victor:
https://sovryn.atlassian.net/browse/SOV-931


soulBit (approved):

packages/ui/src/2_molecules/VerticalTabs/components/VerticalTabItem.tsx
```diff
@@ -22,8 +23,12 @@ export const VerticalTabItem: FC<VerticalTabsItemButtonProps> = ({
       data-active={active}
       onClick={onClick}
     >
-      <p className={styles.label}>{label}</p>
-      {infoText && <small className={styles.info}>{infoText}</small>}
+      <Paragraph size={ParagraphSize.base}>{label}</Paragraph>
```
tiltom, line 26: This is probably just an outdated design, but `label` has `font-size: 1rem` in the design and `0.875rem` in the implementation.

image

packages/ui/src/2_molecules/VerticalTabs/components/VerticalTabItem.tsx
```diff
@@ -22,8 +23,12 @@ export const VerticalTabItem: FC<VerticalTabsItemButtonProps> = ({
...
+      {infoText && (
+        <Paragraph size={ParagraphSize.small} className={styles.info}>
```
tiltom:
```suggestion
        <Paragraph className={styles.info}>
```
`ParagraphSize.small` is default `size`.

tiltom (changes requested):

packages/ui/src/2_molecules/VerticalTabs/components/VerticalTabItem.tsx
soulBit: Since we now have the styling in VerticalTabs matching other usages of these components, we should ask Gilad to update the styling on figma to match, or use the larger text sizes here - I'll ask him to confirm

creed-victor: yeah, i think it should be as Paragraph component  gives to make all paragraphs are in the same size. So current one should be correct one

tiltom (approved):

PR #76, opened 2022-11-10
SOV-817: blockchain rpc fallback

creed-victor:
https://sovryn.atlassian.net/browse/SOV-817

Ignore `packages/ethers-provider/src/lib/fallback-provider.ts` on preview, it's file taken from node_modules with small fix. Will be removed when ethers will merge and release their fix.

### DIFF
--- a/packages/ethers-provider/package.json
+++ b/packages/ethers-provider/package.json
@@ -29,6 +29,14 @@
     "rxjs": "7.5.6"
   },
   "peerDependencies": {
-    "ethers": "5.7"
+    "ethers": "*",
+    "@ethersproject/providers": "*",
+    "@ethersproject/abstract-provider": "*",
+    "@ethersproject/bignumber": "*",
+    "@ethersproject/bytes": "*",
+    "@ethersproject/properties": "*",
+    "@ethersproject/random": "*",
+    "@ethersproject/web": "*",
+    "@ethersproject/logger": "*"
   }
 }
--- a/packages/ethers-provider/src/provider.ts
+++ b/packages/ethers-provider/src/provider.ts
@@ -32,7 +32,7 @@ export function getProvider(
     throw new Error('No chain found');
   }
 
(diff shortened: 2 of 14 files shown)


packages/ethers-provider/package.json
```diff
@@ -29,6 +29,14 @@
     "rxjs": "7.5.6"
   },
   "peerDependencies": {
```
tiltom, line 31: Why don't you use specific versions of the packages?

packages/ethers-provider/src/provider.ts
```diff
@@ -33,13 +32,22 @@ export function getProvider(
     throw new Error('No chain found');
   }
 
-  if (!ethersProviders[chain.rpcUrl]) {
-    ethersProviders[chain.rpcUrl] = new providers.StaticJsonRpcProvider(
-      chain.providerConnectionInfo && chain.providerConnectionInfo.url
-        ? chain.providerConnectionInfo
-        : chain.rpcUrl,
-    );
+  const rpc = chain.rpcUrl === 'string' ? chain.rpcUrl : chain.rpcUrl[0];
```
tiltom:
```suggestion
  const rpc = typeof chain.rpcUrl === 'string' ? chain.rpcUrl : chain.rpcUrl[0];
```

tiltom (changes requested):

packages/ethers-provider/package.json
creed-victor: It's peer dependencies that are actually required on the app itself, not this lib.
I had to add them to the package.lock for `lib/fallback-provider.ts` file to be bundled using rollup cli, but these dependencies must be ignored, otherwise it will not work.

Both peer dependencies and fallback provider file is supposed to be removed later, once ethers will merge their fix to their repo

tiltom (approved):

Rickk137:
Screen Shot 2022-11-11 at 2 18 52 PM
Seems like the provider is making unnecessary call although "mycryptoapi" url is working it makes a call to cloudflare as well which is not needed

Rickk137 (approved):

PR #104, opened 2022-11-28
Sov 1009 table pagination component

Rickk137:
https://sovryn.atlassian.net/browse/SOV-1009


packages/ui/src/2_molecules/AddressTablePagination/AddressTablePagination.module.css
```diff
@@ -1,10 +1,23 @@
 .AddressTablePagination {
```
tiltom:
```suggestion
.paginationWrapper {
```

packages/ui/src/2_molecules/AddressTablePagination/AddressTablePagination.module.css
```diff
@@ -1,10 +1,23 @@
...
-    @apply flex justify-center;
+  @apply flex items-center;
 
-    & > button {
-        @apply mx-8 bg-transparent border-none outline-none cursor-pointer transition-opacity;
-        &:disabled {
-            @apply opacity-50 cursor-not-allowed;
-        }
+  & > .action {
```
tiltom: Why do you use this? You just create unnecessary nesting with this and it's not needed as you there is no need to be afraid of naming collision. Adjust all of the selectors in this file.

packages/ui/src/2_molecules/AddressTablePagination/AddressTablePagination.module.css
```diff
@@ -1,10 +1,23 @@
...
+      @apply opacity-50 cursor-not-allowed;
+    }
+  }
+
+  * > .doubleRight {
+    @apply rotate-180;
+  }
+
+  & > .page {
+    @apply h-6 w-6 rounded mx-1 flex justify-center items-center hover:bg-gray-70 transition-colors;
```
tiltom:
```suggestion
    @apply h-[1.625rem] w-6 rounded mx-1 flex justify-center items-center hover:bg-gray-70 transition-colors;
```
Height should be `26px` according to the design.

packages/ui/src/2_molecules/AddressTablePagination/AddressTablePagination.tsx
```diff
@@ -1,51 +1,104 @@
...
+  previousPage: () => void;
+  gotoPage: (page: number) => void;
+  length?: number;
+  hideLeftArrow?: boolean;
+  hideRightArrow?: boolean;
 };
 
-const DEFAULT_ITEMS_PER_PAGE = 5;
-
 export const AddressTablePagination: FC<AddressTablePaginationProps> = ({
```
tiltom: First of all, this is no longer `AddressTablePagination` but a generic pagination component so it should have an appropriate name, e.g. `Pagination` or `TablePagination`.

packages/ui/src/2_molecules/AddressTablePagination/AddressTablePagination.stories.tsx
```diff
@@ -2,25 +2,81 @@ import { Story } from '@storybook/react';
...
 
 export default {
   title: 'Molecule/AddressTablePagination',
   component: AddressTablePagination,
 };
 
-const Template: Story<ComponentProps<typeof AddressTablePagination>> = args => (
-  <>
+const Template: Story<ComponentProps<typeof AddressTablePagination>> = args => {
+  const {
```
tiltom: This is a wrong composition - why would you need to call `usePaginate` hook in a parent component and then send the result of it as `props` to pagination? You can call the hook directly in the component and the whole component's interface will be much cleaner.

Send just the props that are truly needed from a parent component.

packages/ui/src/2_molecules/AddressTablePagination/AddressTablePagination.stories.tsx
```diff
@@ -2,25 +2,81 @@ import { Story } from '@storybook/react';
...
 export const Basic = Template.bind({});
 Basic.args = {
   className: '',
   dataLayoutId: '',
-  itemsPerPage: 5,
+  pageSize: 6,
+  length: 22,
+};
+
+const _NoLength: Story<
```
tiltom: Misleading, this is not a pagination with "no length", this is a pagination without "total pages count". Rename it.

packages/ui/src/2_molecules/AddressTablePagination/AddressTablePagination.tsx
```diff
@@ -1,51 +1,104 @@
...
+  gotoPage,
+  hideLeftArrow,
+  hideRightArrow,
+}) => (
+  <div
+    {...applyDataAttr(dataLayoutId)}
+    className={classNames(styles.AddressTablePagination, className)}
+  >
+    {!hideLeftArrow && (
+      <button
```
tiltom: You have 4 exactly the same buttons on this page with a few different props, pls extract these buttons into a separate (sub)component.

packages/ui/src/2_molecules/AddressTablePagination/AddressTablePagination.tsx
```diff
@@ -1,51 +1,104 @@
...
+        className={styles.action}
+        type="button"
+        disabled={pageIndex === 0}
+        onClick={() => gotoPage(0)}
```
tiltom: Why you don't have data attributes on the first and last page buttons and only on the next and previous?

packages/ui/src/hooks/usePaginate.ts
```diff
@@ -0,0 +1,34 @@
+import { useState } from 'react';
+
+const DEFAULT_ITEMS_PER_PAGE = 5;
+
+export const usePaginate = (
```
tiltom: If you want this as a hook (which I am not sure is necessary but I am not strongly opposed), you can, but you need to create a `hooks` folder in the component's folder because this is not some generic hook that can (or will be) used anywhere else.

packages/ui/src/hooks/usePaginate.ts
```diff
@@ -0,0 +1,34 @@
...
+  pageSize = DEFAULT_ITEMS_PER_PAGE,
+  listLength?: number,
```
tiltom:
```suggestion
  itemsPerPage = DEFAULT_ITEMS_PER_PAGE,
  totalItems?: number,
```

packages/ui/src/hooks/usePaginate.ts
```diff
@@ -0,0 +1,34 @@
...
+) => {
+  const [activePage, setActivePage] = useState(0);
+  const paginate = <T>(arr: T[]) => {
+    const startIndex = activePage * pageSize;
+    const endIndex =
+      arr?.length && startIndex + pageSize > arr?.length
+        ? arr?.length
+        : startIndex + pageSize;
+    return arr?.slice(startIndex, endIndex);
+  };
```
tiltom:
```suggestion

```
This is wrong and it doesn't belong here - you don't want to perform the actual pagination here, the only thing you want is to retrieve the active page so you can pass it to your graphql query.

packages/ui/src/hooks/usePaginate.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  const pageCount =
```
tiltom:
```suggestion
  const totalPages =
```
Use `useMemo`.

packages/ui/src/hooks/usePaginate.ts
```diff
@@ -0,0 +1,34 @@
...
+    listLength !== undefined ? Math.ceil(listLength / pageSize) : undefined;
+
+  return {
+    activePage,
+    setActivePage,
```
tiltom:
```suggestion

```
Why are you exposing setActivePage? You don't use it anywhere and you shouldn't use it anywhere.

packages/ui/src/hooks/usePaginate.ts
```diff
@@ -0,0 +1,34 @@
...
+    paginate,
+    canNextPage: pageCount === undefined || activePage < pageCount - 1,
```
tiltom: Naming for `canNextPage` and `canPreviousPage` is weird and you always use it with negation (e.g. !canNextPage) to find out if a button is `disabled`. Just name it `isNextButtonDisabled` and `isPreviousButtonDisabled` and it will be much cleaner.

Also, use `useMemo`.

packages/ui/src/hooks/usePaginate.ts
```diff
@@ -0,0 +1,34 @@
...
+    canPreviousPage: activePage > 0,
+    gotoPage: setActivePage,
+    nextPage: () => setActivePage(activePage + 1),
```
tiltom:
```suggestion
    onNextPage: () => setActivePage(activePage + 1),
```
Use `useCallback`.

packages/ui/src/hooks/usePaginate.ts
```diff
@@ -0,0 +1,34 @@
...
+    pageCount,
+    pageSize,
+    previousPage: () => setActivePage(activePage - 1),
```
tiltom:
```suggestion
    onPreviousPage: () => setActivePage(activePage - 1),
```
Use `useCallback`.

packages/ui/src/hooks/usePaginate.ts
```diff
@@ -0,0 +1,34 @@
...
+  return {
+    activePage,
+    setActivePage,
+    paginate,
+    canNextPage: pageCount === undefined || activePage < pageCount - 1,
+    canPreviousPage: activePage > 0,
+    gotoPage: setActivePage,
+    nextPage: () => setActivePage(activePage + 1),
+    pageCount,
+    pageSize,
```
tiltom:
```suggestion

```
Why do you return it if it's sent by a pagination component?

packages/ui/src/hooks/usePaginate.ts
```diff
@@ -0,0 +1,34 @@
...
+  const pageCount =
+    listLength !== undefined ? Math.ceil(listLength / pageSize) : undefined;
+
+  return {
+    activePage,
+    setActivePage,
+    paginate,
+    canNextPage: pageCount === undefined || activePage < pageCount - 1,
+    canPreviousPage: activePage > 0,
+    gotoPage: setActivePage,
```
tiltom:
```suggestion
    onPageChange: setActivePage,
```

tiltom (changes requested):

packages/ui/src/2_molecules/Pagination/components/PaginationButton/PaginationButton.module.css
```diff
@@ -0,0 +1,7 @@
+.button {
+  @apply bg-transparent border-none outline-none cursor-pointer transition-opacity mx-1;
```
pietro-maximoff:
```suggestion
  @apply text-gray-10 bg-transparent border-none outline-none cursor-pointer transition-opacity mx-1;
```

pietro-maximoff (changes requested):

packages/ui/src/2_molecules/Pagination/Pagination.stories.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+import { Pagination } from './Pagination';
+
+export default {
+  title: 'Molecule/Pagination',
+  component: Pagination,
+};
+
+const Template: Story<ComponentProps<typeof Pagination>> = args => {
+  const [page, setPage] = useState(0);
```
tiltom, line 13: I don't think that's a good approach to set pagination state from a parent component, do you have a specific use case for this? I think it should have an internal state and in case you need to override it, just set `page` prop.

packages/ui/src/2_molecules/Pagination/Pagination.test.tsx
```diff
@@ -0,0 +1,31 @@
...
+describe('Pagination', () => {
+  it('renders Pagination', () => {
+    const gotoPage = jest.fn();
+    const { findAllByRole } = render(
+      <Pagination setPage={gotoPage} page={0} totalItems={20} />,
+    );
+    expect(findAllByRole('svg[data-icon="arrow-back"]')).toBeDefined();
+  });
+
+  it('does not allow to click Previous without clicking Next first', () => {
```
tiltom: This label is misleading because you don't test what it says, you just click "Next". Rename it and add a test that actually tests the described behaviour.

packages/ui/src/2_molecules/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,101 @@
...
+import styles from './Pagination.module.css';
+import { PaginationButton } from './components/PaginationButton/PaginationButton';
+import { usePaginate } from './hooks/usePaginate';
+
+type PaginationProps = {
+  className?: string;
+  dataLayoutId?: string;
+  pageSize?: number;
+  page: number;
+  setPage: (page: number) => void;
```
tiltom: As mentioned earlier, this is not correct but I think that you can rename this to `onChange` and it should be called every time we have a page change (and it should be optional ofc).

packages/ui/src/2_molecules/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+import { IconNames } from '../../1_atoms';
+import { applyDataAttr } from '../../utils';
+import styles from './Pagination.module.css';
+import { PaginationButton } from './components/PaginationButton/PaginationButton';
+import { usePaginate } from './hooks/usePaginate';
+
+type PaginationProps = {
+  className?: string;
+  dataLayoutId?: string;
```
tiltom:
```suggestion
  dataAttribute?: string;
```

packages/ui/src/2_molecules/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,101 @@
...
+  pageSize?: number;
```
tiltom:
```suggestion
  itemsPerPage?: number;
```

packages/ui/src/2_molecules/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,101 @@
...
+  page: number;
+  setPage: (page: number) => void;
+  totalItems?: number;
+  hideLeftArrow?: boolean;
+  hideRightArrow?: boolean;
```
tiltom:
```suggestion
  hideFirstPageButton?: boolean;
  hideLastPageButton?: boolean;
```

packages/ui/src/2_molecules/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+  return (
+    <div
+      {...applyDataAttr(dataLayoutId)}
+      className={classNames(styles.paginationWrapper, className)}
+    >
+      {!hideLeftArrow && (
+        <PaginationButton
+          disabled={page === 0}
+          onClick={() => setPage(0)}
```
tiltom: Use `useCallback`.

packages/ui/src/2_molecules/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,101 @@
...
+      <PaginationButton
+        disabled={isNextButtonDisabled}
+        onClick={onNextPage}
+        dataLayoutId={`${dataLayoutId}-next`}
+        icon={IconNames.ARROW_FORWARD}
+      />
+      {totalPages !== undefined && !hideRightArrow && (
+        <PaginationButton
+          disabled={page + 1 === totalPages}
+          onClick={() => setPage(totalPages - 1)}
```
tiltom: Use `useCallback`.

packages/ui/src/2_molecules/Pagination/components/PaginationButton/PaginationButton.tsx
```diff
@@ -0,0 +1,31 @@
+import React, { FC, MouseEventHandler } from 'react';
+
+import { Icon, IconType } from '../../../../1_atoms';
+import { applyDataAttr } from '../../../../utils';
+import styles from './PaginationButton.module.css';
+
+type PaginationButtonProps = {
+  dataLayoutId?: string;
```
tiltom:
```suggestion
  dataAttribute?: string;
```

tiltom (changes requested):

packages/ui/src/2_molecules/Pagination/Pagination.stories.tsx
Rickk137: the data is fetched in the parent component. that's why it needs the page number. also you might need to reset it if the filters has changed

packages/ui/src/2_molecules/Pagination/hooks/usePaginate.ts
```diff
@@ -0,0 +1,46 @@
+import { useCallback, useMemo } from 'react';
+
+export const usePaginate = (
+  pageIndex: number,
+  setPageIndex: (page: number) => void,
+  itemsPerPage: number,
+  listLength?: number,
```
tiltom:
```suggestion
  page: number,
  onChange: (page: number) => void,
  itemsPerPage: number,
  totalItems?: number,
```

packages/ui/src/2_molecules/Pagination/Pagination.test.tsx
```diff
@@ -0,0 +1,30 @@
+import { render } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+
+import React from 'react';
+
+import { Pagination } from './Pagination';
+
+describe('Pagination', () => {
```
tiltom, line 8: Add also a test for "Next" button.

tiltom (changes requested):

tiltom (approved):

pietro-maximoff (approved):
checked, looks good

LunaSov (approved):

PR #105, opened 2022-11-28
SOV-1087: add getTokenInfoByAddress

Rickk137:
https://sovryn.atlassian.net/browse/SOV-1087

### DIFF
--- a/apps/frontend/src/app/3_organisms/TransactionStepDialog/components/TransactionStep/TransactionStep.tsx
+++ b/apps/frontend/src/app/3_organisms/TransactionStepDialog/components/TransactionStep/TransactionStep.tsx
@@ -1,9 +1,10 @@
-import React, { FC, useCallback, useMemo, useState } from 'react';
+import React, { FC, useCallback, useEffect, useMemo, useState } from 'react';
 
 import classNames from 'classnames';
 import { formatUnits, parseUnits } from 'ethers/lib/utils';
 import { bignumber } from 'mathjs';
 
+import { getTokenInfo, TokenBaseInfo } from '@sovryn/contracts';
 import {
   Accordion,
   AmountInput,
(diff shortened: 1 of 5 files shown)


packages/contracts/src/utils/global.ts
```diff
@@ -165,3 +165,10 @@ export const getTokenDetailsData = async (
 
   return tokenDetails;
 };
+
+export const getTokenInfoByAddress = async (
```
tiltom: I think you should do it the same way I told you before - to unify existing `getTokenDetailsData` function with the new one and use `address` as a parameter for the new function and `tokenName and chainId` for the current function.

I'd prefer not to obtain only `TokenBaseInfo` because if you return the `TokenDetailsData` instead, you'll cover all of the possible use cases we may have in the future.

packages/contracts/src/utils/getters.ts
```diff
@@ -26,3 +31,7 @@ export const getTokenDetails = async (
   token: SupportedTokens,
   chain: ChainId = ChainIds.RSK_MAINNET,
 ): Promise<TokenDetailsData> => getTokenDetailsData(token, chain);
+
+export const getTokenInfo = async (
```
tiltom:
```suggestion
export const getTokenDetailsByAddress = async (
```

apps/frontend/src/app/3_organisms/TransactionStepDialog/components/TransactionStep/TransactionStep.tsx
```diff
@@ -48,15 +48,12 @@ export const TransactionStep: FC<TransactionStepProps> = ({
...
-          token.address.toLowerCase() ===
-          transaction.contract.address.toLowerCase(),
-      ),
-    [transaction.contract.address],
-  );
+  const [token, setToken] = useState<TokenBaseInfo | undefined>();
+
+  useEffect(() => {
+    (() => getTokenInfo(transaction.contract.address).then(setToken))();
+    //eslint-disable-next-line
```
tiltom: Why?

tiltom (dismissed):

soulBit (approved):
Verified review comments have been addressed, and made two small fixes (already discussed with Rick)

PR #109, opened 2022-11-30
SOV-1086: feat: helper functions for working with ethers bignumbers

creed-victor:
https://sovryn.atlassian.net/browse/SOV-1086

### DIFF
--- a/apps/frontend/src/utils/math.test.ts
+++ b/apps/frontend/src/utils/math.test.ts
@@ -9,8 +9,7 @@ const gasLimit = 47254;
 describe('utils/math.ts', () => {
   describe('toWei()', () => {
     it('converts floating numbers to wei', () => {
-      const value = toWei(gasPrice).toString();
-      expect(value).toEqual('65164000000000000');
+      expect(toWei(gasPrice).toString()).toEqual('65164000000000000');
     });
 
     it('converts gwei floating numbers to wei', () => {
@@ -21,8 +20,7 @@ describe('utils/math.ts', () => {
     });
 
     it('converts full number to wei', () => {
-      const value = toWei(47254).toString();
-      expect(value).toEqual('47254000000000000000000');
+      expect(toWei(47254).toString()).toEqual('47254000000000000000000');
     });
 
     it('converts 1 wei to 1**0', () => {
@@ -130,7 +128,6 @@ describe('ethers/lib/utils', () => {
   });
 
   it('adds comma as thousands separator', () => {
-    const value = commify(123456.789);
-    expect(value).toBe('123,456.789');
+    expect(commify(123456.789)).toBe('123,456.789');
   });
 });
--- a/apps/frontend/src/utils/math.test.ts
+++ b/apps/frontend/src/utils/math.test.ts
@@ -1,0 +1,87 @@
+import { commify, formatUnits, parseUnits } from 'ethers/lib/utils';
+
+import { fromWei, toWei } from './math';
+
+const gasPrice = 0.065164;
+const gasLimit = 47254;
+
+describe('utils/math.ts', () => {
+  describe('toWei()', () => {
+    it('converts floating numbers to wei', () => {
+      const value = toWei(gasPrice).toString();
+      expect(value).toEqual('65164000000000000');
+    });
+
+    it('converts gwei floating numbers to wei', () => {
+      const value = toWei(gasPrice)
+        .div(10 ** 9)
+        .toString();
+      expect(value).toEqual('65164000');
+    });
+
+    it('converts full number to wei', () => {
+      const value = toWei(47254).toString();
+      expect(value).toEqual('47254000000000000000000');
+    });
+
+    it('fails for unexpected value', () => {
+      expect(() => toWei('error')).toThrowError(
+        'Invalid BigNumberish value: error',
+      );
+    });
+  });
+
+  describe('fromWei()', () => {
+    it('convert number to rbtc', () => {
+      const value = fromWei(47254);
+      expect(value).toEqual('0.000000000000047254');
+    });
+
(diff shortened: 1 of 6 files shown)


apps/frontend/src/utils/math.test.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  it('returns gas in gwei', () => {
+    const value = formatUnits(
+      parseUnits(gasPrice.toString(), 'gwei').mul(gasLimit),
+      'gwei',
+    );
+    expect(value).toBe('3079.259656');
+  });
+
+  it('returns gas in gwei', () => {
```
tiltom: You have 2 identical tests.

apps/frontend/src/utils/math.test.ts
```diff
@@ -0,0 +1,87 @@
...
+    const value = formatUnits(
+      parseUnits(gasPrice.toString(), 'gwei').mul(gasLimit),
+      'gwei',
+    );
+    expect(value).toBe('3079.259656');
+  });
+
+  it('adds comma as thousand separator', () => {
```
tiltom:
```suggestion
  it('adds comma as thousands separator', () => {
```

apps/frontend/src/utils/math.ts
```diff
@@ -0,0 +1,29 @@
+import { isBigNumberish } from '@ethersproject/bignumber/lib/bignumber';
+
+import { BigNumberish, BigNumber } from 'ethers';
+import { formatUnits, parseUnits } from 'ethers/lib/utils';
+
+// helper function to convert any type of ethers value to wei.
+export const toWei = (value: BigNumberish): BigNumber => {
+  if (isBigNumberish(value)) {
+    return BigNumber.from(value).mul(BigNumber.from(10).pow(18));
```
tiltom: I think that this depends on a number of decimals, right? I know that 18 is the most common but what if we have a token with a different amount of decimals?

tiltom (changes requested):

apps/frontend/src/utils/math.ts
creed-victor: good point, will add argument for overriding decimals points.

apps/frontend/src/utils/math.test.ts
```diff
@@ -0,0 +1,136 @@
...
+import { commify, formatUnits, parseUnits } from 'ethers/lib/utils';
+
+import { fromWei, fromWeiFixed, toWei } from './math';
+
+const gasPrice = 0.065164;
+const gasLimit = 47254;
+
+describe('utils/math.ts', () => {
+  describe('toWei()', () => {
+    it('converts floating numbers to wei', () => {
```
tiltom, line 11: Rewrite to one-liner pls.

apps/frontend/src/utils/math.test.ts
```diff
@@ -0,0 +1,136 @@
...
+    });
+
+    it('converts gwei floating numbers to wei', () => {
+      const value = toWei(gasPrice)
+        .div(10 ** 9)
+        .toString();
+      expect(value).toEqual('65164000');
+    });
+
+    it('converts full number to wei', () => {
```
tiltom, line 22: Rewrite to one-liner pls.

apps/frontend/src/utils/math.test.ts
```diff
@@ -0,0 +1,136 @@
...
+
+  it('returns gas in gwei', () => {
+    const value = formatUnits(
+      parseUnits(gasPrice.toString(), 'gwei').mul(gasLimit),
+      'gwei',
+    );
+    expect(value).toBe('3079.259656');
+  });
+
+  it('adds comma as thousands separator', () => {
```
tiltom, line 130: Rewrite to one-liner pls.

apps/frontend/src/utils/math.ts
```diff
@@ -0,0 +1,50 @@
+import { isBigNumberish } from '@ethersproject/bignumber/lib/bignumber';
+
+import { BigNumberish, BigNumber } from 'ethers';
+import { formatUnits, parseUnits } from 'ethers/lib/utils';
+
+const DEFAULT_UNIT = 18;
+const DEFAULT_DECIMALS = 6;
+
+const unitNames = ['wei', 'kwei', 'mwei', 'gwei', 'szabo', 'finney', 'ether'];
```
tiltom, line 9:
```suggestion
const units = {
  wei: 0,
  kwei: 3,
  mwei: 6,
  gwei: 9,
  szabo: 12,
  finney: 15,
  ether: 18,
};
```
I'd rather have an object with mapping than just an array.

apps/frontend/src/utils/math.ts
```diff
@@ -0,0 +1,50 @@
...
+export const toWei = (
+  value: BigNumberish,
+  unitName: BigNumberish = DEFAULT_UNIT,
+): BigNumber => {
+  if (isBigNumberish(value)) {
+    if (typeof unitName === 'string') {
+      const index = unitNames.indexOf(unitName);
+      if (index !== -1) {
+        unitName = 3 * index;
+      }
```
tiltom, line 21:
```suggestion
unitName = units[unitName];
```
I think this is a better approach compared to multiplying an index from the `unitNames` array with a seemingly random number (`3`).

tiltom (changes requested):

apps/frontend/src/utils/math.ts
creed-victor: ethers bignumber accepts units as it's defined here so I prefer to use it same way to be fully comply with it.

tiltom: ok then, let's leave it that way

tiltom (approved):

soulBit (approved):

PR #110, opened 2022-12-01
SOV-1085: Update AmountInput styles

tiltom:
Resolves https://sovryn.atlassian.net/browse/SOV-1085

### DIFF
--- a/packages/ui/src/2_molecules/AmountInput/AmountInput.module.css
+++ b/packages/ui/src/2_molecules/AmountInput/AmountInput.module.css
@@ -3,7 +3,7 @@
 }
 
 .container {
-  @apply text-xs relative flex text-gray-20 flex-grow rounded bg-gray-70 items-center px-2 outline outline-1 outline-gray-50;
+  @apply text-xs relative flex text-gray-20 flex-grow rounded bg-gray-60 items-center px-2 outline outline-1 outline-gray-50;
 
   &:hover {
     @apply outline-gray-40;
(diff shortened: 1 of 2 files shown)


packages/ui/src/2_molecules/AmountInput/AmountInput.module.css
```diff
@@ -43,11 +43,19 @@
 }
 
 .invalid {
-  @apply outline outline-1 outline-error-light bg-gray-60;
+  @apply outline outline-1 outline-error-light bg-gray-70;
```
soulBit:
```suggestion
  @apply outline outline-1 outline-error-light bg-gray-60;
```
Figma shows color as #333845 for invalid state, but it's #2c303b in storybook

packages/ui/src/2_molecules/AmountInput/AmountInput.module.css
```diff
@@ -43,11 +43,19 @@
...
+
+  &:hover {
+    @apply outline-error-light;
```
soulBit, line 49: Invalid border color seems to take precedence here as it doesn't change when hovering in storybook, but I don't see this in Figma so I think this can just be removed.

soulBit (changes requested):
Added changeset as discussed

packages/ui/src/2_molecules/AmountInput/AmountInput.module.css
tiltom: This was actually requested by Gilad because without this, you could hover over invalid/focused states and you'd see hover effect that shouldn't be there.

soulBit (approved):

pietro-maximoff (approved):

PR #117, opened 2022-12-06
SOV-1090: Zero: LOC getting started popup

pietro-maximoff:
https://sovryn.atlassian.net/browse/SOV-1090


.changeset/tiny-seas-approve.md
```diff
@@ -0,0 +1,5 @@
+---
+"frontend": patch
+---
+
+Zero: LOC getting started popup
```
tiltom:
```suggestion
SOV-1090: Zero - LOC getting started popup
```

apps/frontend/src/app/2_molecules/DebugContent.tsx
```diff
@@ -151,6 +153,26 @@ export const DebugContent = () => {
...
+        <Button text="Open Getting Started Popup" onClick={togglePopup} />
+
+        {isPopupOpen && (
+          <>
+            <Dialog
+              isOpen={isPopupOpen}
+              onClose={togglePopup}
+              className="p-10 text-center"
+            >
+              <p>Placeholder for ZERO LOC transaction dialog</p>
```
tiltom: Why you just don't show `GettingStartedPopup`? I don't understand why is this necessary and I don't see it in the requirements either.

apps/frontend/src/app/3_organisms/GettingStartedPopup/GettingStartedPopup.module.css
```diff
@@ -0,0 +1,71 @@
+.dialog {
+  @apply py-7 px-4;
+
+  & > h1, & > h2 {
+    @apply leading-5;
```
tiltom: I don't think this is correct, one advantage of a UI components library is that we have a set of components ready to use and since you use `Heading` component, it should use `line-height` defined there. Consult it with Gilad if you can use default `line-height` from `Heading` instead.

apps/frontend/src/app/3_organisms/GettingStartedPopup/GettingStartedPopup.module.css
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  & > h1 {
```
tiltom:
```suggestion
  .title {
```
I don't like this approach, the naming should be descriptive and not just a copy of the current CSS structure. If we decide in the future to change `h1` and `h2` to `h3` and `h4`, styles would just stop working and I can imagine it wouldn't be a pleasant debugging.

apps/frontend/src/app/3_organisms/GettingStartedPopup/GettingStartedPopup.module.css
```diff
@@ -0,0 +1,71 @@
...
+    @apply mb-3 font-normal;
+  }
+
+  & > h2 {
```
tiltom:
```suggestion
  .description {
```

apps/frontend/src/app/3_organisms/GettingStartedPopup/GettingStartedPopup.module.css
```diff
@@ -0,0 +1,71 @@
...
+    @apply leading-5 mb-2;
+  }
+
+  & > h2 {
+    @apply text-white;
+  }
+
+  & > h3 {
+    @apply text-gray-10 font-medium;
+  }
```
tiltom: The exact same comments I had regarding `title` and `description`.

apps/frontend/src/app/3_organisms/GettingStartedPopup/GettingStartedPopup.module.css
```diff
@@ -0,0 +1,71 @@
...
+
+.content {
+  @apply flex-col py-5 px-3 rounded bg-gray-80 mb-5;
+
+  @screen md {
+    @apply mb-[4.5rem] py-7 px-5;
+  }
+}
+
+.box {
```
tiltom:
```suggestion
.itemWrapper {
```

apps/frontend/src/app/3_organisms/GettingStartedPopup/GettingStartedPopup.module.css
```diff
@@ -0,0 +1,71 @@
...
+  &:last-child {
+    @apply border-b-0 mb-0;
+  }
+
+  @screen md {
+    @apply pb-6;
+  }
+}
+
+.boxContent {
```
tiltom:
```suggestion
.texts {
```

apps/frontend/src/app/3_organisms/GettingStartedPopup/GettingStartedPopup.tsx
```diff
@@ -0,0 +1,139 @@
...
+  Link,
+  Pagination,
+} from '@sovryn/ui';
+
+import { useIsMobile } from '../../../hooks/useIsMobile';
+import { translations } from '../../../locales/i18n';
+import { sovrynWikiLinks } from '../../../utils/constants';
+import styles from './GettingStartedPopup.module.css';
+
+const DEFAULT_PAGE_SIZE = 1;
```
tiltom:
```suggestion
const pageSize = 1;
```
You don't offer any option to override `pageSize` so the naming is a bit misleading. Also, it's not a global constant but a local one so I think that camelCase should be fine (but it can be `PAGE_SIZE` if you agree on that with @soulBit).

apps/frontend/src/app/3_organisms/GettingStartedPopup/GettingStartedPopup.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+export const GettingStartedPopup: FC = () => {
+  const { t } = useTranslation();
+  const { isMobile } = useIsMobile();
+  const [page, setPage] = useState(0);
+  const [isOpen, setIsOpen] = useState(false);
```
tiltom: I think you should have 2 props - `isOpen` and `onConfirm` and it's not something that should be in the component itself.

apps/frontend/src/app/3_organisms/GettingStartedPopup/GettingStartedPopup.tsx
```diff
@@ -0,0 +1,139 @@
...
+  );
+
+  const items = useMemo(
+    () => (isMobile ? data.slice(page, pageOffset) : data),
+    [data, pageOffset, isMobile, page],
+  );
+
+  const onClick = useCallback(() => {
+    if (checked) {
+      reactLocalStorage.set('gettingStartedPopup', 'false');
```
tiltom: Extract `gettingStartedPopup` to a variable (e.g. `localStorageKey`) and use that instead.

tiltom (changes requested):

apps/frontend/src/app/2_molecules/DebugContent.tsx
pietro-maximoff: it's just for example, and its not necessary

apps/frontend/src/app/3_organisms/GettingStartedPopup/GettingStartedPopup.module.css
pietro-maximoff: confirmed with Gilad

apps/frontend/src/app/3_organisms/GettingStartedPopup/GettingStartedPopup.tsx
pietro-maximoff: @tiltom i'm not sure about it, because it wasn't in the specs and if i will have these props, then i probably should move the logic with localStorage outside of the component as well, because if i send for example "true" in the isOpen prop and then inside the component, i detect that i have localStorageKey - it means that i shouldn't show it - it won't make any sense) but if you think that we strongly should have these props, then pls describe in more detail what logic we should have here

tiltom: I think that you wouldn't need to move the local storage logic outside the component - you would just check in an `effect` if `isOpen` is `true` and `localStorageKey` exists and if so, you would call `onConfirm` (which will set `isOpen` to `false` in a parent component). Or do you think it won't work that way?

pietro-maximoff: @tiltom i managed it to work in this way, you can check and I replaced `useEffect` with `useLayoutEfect`, because with `useEffect` this dialog blinked before hiding

tiltom (approved):

soulBit (approved):

PR #158, opened 2023-01-17
Sov 1287 zero stability pool page implementation

Rickk137:
https://sovryn.atlassian.net/browse/SOV-1287


.env
```diff
@@ -0,0 +1,7 @@
+REACT_APP_GOOGLE_ANALYTICS=GTM-PX36D3G
+
+REACT_APP_GRAPH_RSK=https://subgraph.test.sovryn.app/subgraphs/name/DistributedCollective/sovryn-subgraph
+
+REACT_APP_GRAPH_ZERO=https://zero-subgraph.test.sovryn.app/subgraphs/name/DistributedCollective/zero-subgraph
```
soulBit: Remove this file, I don't think this should be here

apps/frontend/src/app/3_organisms/TransactionStepDialog/components/TransactionStep/TransactionStep.tsx
```diff
@@ -71,7 +74,8 @@ export const TransactionStep: FC<TransactionStepProps> = ({
             ? transaction.args[1]
             : undefined,
         gasPrice,
-        gasLimit: gasLimit.toString(),
+        //Increase gas limit by 20% to make sure tx won't fail for this reason
```
soulBit:
```suggestion
        //TODO: replace with default gas limit - increase gas limit by 20% to make sure tx won't fail
```

Update comment here so we can find it easily later when we replace these with suggested default gas limits, and extract calculation to separate section (same thing for other instance)

apps/frontend/src/app/3_organisms/TransactionStepDialog/components/TransactionSteps/TransactionSteps.tsx
```diff
@@ -153,12 +161,8 @@ export const TransactionSteps: FC<TransactionStepsProps> = ({
...
       )}
 
-      {onSuccess && transactions.length === step && (
-        <Button
-          text="Done"
-          onClick={onSuccess}
-          className="w-full mt-7"
-        ></Button>
+      {onClose && transactions.length === step && (
+        <Button text="Done" onClick={onClose} className="w-full mt-7"></Button>
```
soulBit: Add localisation for the button text

apps/frontend/src/app/5_pages/EarnPage/EarnPage.tsx
```diff
@@ -1,12 +1,318 @@
...
+            type={TabType.secondary}
+            items={actions}
+            onChange={setIndex}
+            index={index}
+          />
+
+          <button
+            onClick={onMaximumAmountClick}
+            className="text-xs font-medium underline whitespace-nowrap"
+            {...applyDataAttr('max-button')}
```
soulBit:
```suggestion
            {...applyDataAttr('earn-max-button')}
```

Prefix all the data attributes on this page with `earn-`

apps/frontend/src/app/5_pages/EarnPage/hooks/useHandleStabalityDeposit.ts
```diff
@@ -0,0 +1,97 @@
...
+import { getContract } from '@sovryn/contracts';
+
+import { Transaction } from '../../../3_organisms/TransactionStepDialog/TransactionStepDialog.types';
+import { useTransactionContext } from '../../../../contexts/TransactionContext';
+import { useAccount } from '../../../../hooks/useAccount';
+import { getRskChainId } from '../../../../utils/chain';
+import { toWei } from '../../../../utils/math';
+import { useHandleConversion } from '../../ConvertPage/hooks/useHandleConversion';
+
+export const useHandleStabalityDeposit = (
```
soulBit:
```suggestion
export const useHandleStabilityDeposit = (
```
Rename file too

apps/frontend/src/router.tsx
```diff
@@ -42,6 +42,7 @@ export const router = createBrowserRouter([
       {
         path: '/earn',
         element: <EarnPage />,
+        loader: zeroPageLoader,
```
soulBit: If we are going to use the ethers-provider from this import, please extract the file from `apps\frontend\src\app\5_pages\ZeroPage\loader.ts` to somewhere else more generic - please discuss with team first, I suggest somewhere like `apps\frontend\src\store\EthersProviderLoader.ts` or similar (`store` folder maybe probably isn't quite correct)

.env
```diff
@@ -0,0 +1,7 @@
...
+
+REACT_APP_NETWORK=testnet
```
pietro-maximoff: pls add a new line

apps/frontend/src/app/3_organisms/TransactionStepDialog/components/TransactionStep/TransactionStep.tsx
```diff
@@ -71,7 +74,8 @@ export const TransactionStep: FC<TransactionStepProps> = ({
...
+        gasLimit: gasLimit.mul(15).div(10).toString(),
```
pietro-maximoff: since this calculation is used multiple times, you can store the result in a variable and reuse it instead of recalculating it each time.

apps/frontend/src/app/3_organisms/TransactionStepDialog/components/TransactionSteps/TransactionSteps.tsx
```diff
@@ -39,7 +41,7 @@ export const TransactionSteps: FC<TransactionStepsProps> = ({
           amount: tx.fnName === APPROVAL_FUNCTION ? tx.args[1] : undefined,
           unlimitedAmount: tx.fnName === APPROVAL_FUNCTION ? false : undefined,
           gasPrice,
-          gasLimit: gasLimit.toString(),
+          gasLimit: gasLimit.mul(15).div(10).toString(),
```
pietro-maximoff: since this calculation is used multiple times, you can store the result in a variable and reuse it instead of recalculating it each time.

apps/frontend/src/app/5_pages/EarnPage/EarnPage.tsx
```diff
@@ -1,12 +1,318 @@
...
+  const actions = useMemo(() => {
+    const tabs = [
+      {
+        label: t(commonTranslations.deposit),
+        activeClassName: 'text-primary-20',
+        dataAttribute: 'deposit',
+      },
+    ];
+
+    if (!BigNumber.from(toWei(poolBalance)).isZero()) {
```
pietro-maximoff, line 91: i believe it's better to move this computation to its own useMemo and pass it as a dependency there

apps/frontend/src/app/5_pages/EarnPage/EarnPage.tsx
```diff
@@ -1,12 +1,318 @@
...
+    ).toString();
+  }, [ZUSDInStabilityPool, poolBalance]);
+
+  const isAmountZero = useMemo(() => {
+    return Number(amount) === 0;
+  }, [amount]);
+
+  const newPoolBalance = useMemo(() => {
+    if (isAmountZero) {
+      return 'N/A';
```
pietro-maximoff: use t(translations.common.na) and replace everywhere

pietro-maximoff (changes requested):

soulBit (changes requested):

apps/frontend/src/app/3_organisms/TransactionStepDialog/components/TransactionStep/TransactionStep.tsx
Rickk137: well this is a temporary solution for now. cause we are planing to change gas limits in a separated story

apps/frontend/src/app/5_pages/EarnPage/EarnPage.tsx
Rickk137: well this calculation is only used in here. so even if we move it to a separate useMemo that would be only used here so we are only adding a memory use and gain no performance improvements.

soulBit (changes requested):
Few adjustments to make:
1. hide Deposit button if no existing SP deposit (since they are always depositing)
2. Title is not matching latest Figma: "Earn" -> "Stability Pool"
3. Reset the amount input value when switching between Deposit and Withdraw

apps/frontend/src/router.tsx
soulBit: discussed with @Rickk137 and will be refactored separately

apps/frontend/src/app/3_organisms/TransactionStepDialog/components/TransactionStep/TransactionStep.tsx
soulBit: fine to leave here for now, we'll update it next sprint

apps/frontend/src/app/3_organisms/TransactionStepDialog/components/TransactionSteps/TransactionSteps.tsx
soulBit: fine to leave here for now, we'll update it next sprint

soulBit (changes requested):
image
In case of withdrawals, max amount should be SP balance (in this case it would be 25 ZUSD) - see point 3 in spec

soulBit (approved):

pietro-maximoff (approved):

PR #179, opened 2023-01-26
SOV-1521: deduct fees from collateral

creed-victor:
https://sovryn.atlassian.net/browse/SOV-1521


apps/frontend/src/app/2_molecules/BalanceInput/BalanceInput.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+import { SupportedTokens } from '@sovryn/contracts';
+import { ChainId } from '@sovryn/ethers-provider';
+import { AmountInput, AmountInputProps } from '@sovryn/ui';
+
+import { useMaxAssetBalance } from '../../../hooks/useMaxAssetBalance';
+import { getRskChainId } from '../../../utils/chain';
+import { fromWei } from '../../../utils/math';
+
+const DEFAULT_GAS_LIMIT = 1000000; // 1M wei
```
tiltom: I'd change it to `6 800 000`, see the comment from @john-light : https://sovryn.atlassian.net/browse/SOV-1013?focusedCommentId=11653 . I'd also rename it to `MAX_GAS_LIMIT` and move it somewhere else as it's used in `useMaxAssetBalance`.

@john-light are you ok with us subtracting `gasLimit * gasPrice` if we have a custom gas limit passed to the component and use `DEFAULT_GAS_LIMIT` only as a fallback or do you want us to use it all the time?

apps/frontend/src/app/2_molecules/BalanceInput/index.tsx
```diff
@@ -0,0 +1 @@
+export * from './BalanceInput';
```
tiltom: This file is probably not needed as we won't use it outside of the fronted dapp, right?

apps/frontend/src/hooks/useAsync.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export const useAsync = <T>(asyncFunction: () => Promise<T>) => {
+  const [value, setValue] = useState<T>();
+  // useCallback ensures the below useEffect is not called
+  // on every render, but only if asyncFunction changes.
+  const execute = useCallback(() => {
+    return asyncFunction()
+      .then(setValue)
+      .catch(() => setValue(undefined));
+  }, [asyncFunction]);
```
tiltom:
```suggestion
  const execute = useCallback(
    () =>
      asyncFunction()
        .then(setValue)
        .catch(() => setValue(undefined)),
    [asyncFunction],
  );
```

apps/frontend/src/app/3_organisms/ZeroLocForm/AdjustCreditLine.tsx
```diff
@@ -101,8 +103,10 @@ export const AdjustCreditLine: FC<AdjustCreditLineProps> = ({
 
   const { value: creditWeiBalance } = useAssetBalance(debtToken);
 
-  const { value: maxCollateralWeiAmount } = useAssetBalance(
+  const { value: maxCollateralWeiAmount } = useMaxAssetBalance(
     SupportedTokens.rbtc,
+    getRskChainId(),
+    collateralType === AmountType.Add ? GAS_LIMIT_OPEN_TROVE : 0,
```
tiltom:
```suggestion
    collateralType === AmountType.Add ? GAS_LIMIT_OPEN_TROVE : undefined,
```

apps/frontend/src/app/2_molecules/BalanceInput/BalanceInput.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+type BalanceInputProps = Omit<AmountInputProps, 'maxAmount' | 'unit'> & {
+  token: SupportedTokens;
+  gasLimit?: BigNumberish;
+  chainId?: ChainId;
+};
+
+export const BalanceInput = forwardRef<HTMLInputElement, BalanceInputProps>(
```
tiltom: I don't see a reason for having this component as it just wraps `AmountInput` and sets `maxAmount` prop but in majority of the cases, you will need to have that `maxAmount` value in the parent component and use it for a label/max button as well.

tiltom (changes requested):

apps/frontend/src/app/2_molecules/BalanceInput/BalanceInput.tsx
creed-victor: Having it as a fee in the max amount makes sense and makes it easer for us to handle.

In case there was a plan to use this fee in the actual transaction - noting that 6.8M is overkill to use and should be avoided whenever possible.

cc @john-light

apps/frontend/src/app/2_molecules/BalanceInput/BalanceInput.tsx
creed-victor: You are right, without "MAX" button it not useful at all. Will remove it.

apps/frontend/src/app/3_organisms/ZeroLocForm/AdjustCreditLine.tsx
creed-victor: `0` is correct for this particular case, because gas fee should not be removed for withdrawing collateral (because rbtc balance is taken not from user wallet, but withdrawing from contract)

apps/frontend/src/app/2_molecules/BalanceInput/BalanceInput.tsx
john-light: > are you ok with us subtracting gasLimit * gasPrice if we have a custom gas limit passed to the component and use DEFAULT_GAS_LIMIT only as a fallback or do you want us to use it all the time?

@tiltom 
Can you suggest a scenario where the user is passing a custom gas limit _before_ selecting the amount of their tx?

If you cannot suggest such a scenario, then for the purposes of calculating "max amount" I request that you use the formula I specified in Jira.

> In case there was a plan to use this fee in the actual transaction - noting that 6.8M is overkill to use and should be avoided whenever possible.

@creed-victor 
Yes, unless the tx the user is making really requires 6.8M gas, a more reasonable default gas limit should be used. (There is a different story for defining default gas limits [here](https://sovryn.atlassian.net/browse/SOV-1452).)

creed-victor: Ok, this conversation can be resolved then

apps/frontend/src/hooks/useMaxAssetBalance.ts
```diff
@@ -0,0 +1,43 @@
...
+export const useMaxAssetBalance = (
+  asset: SupportedTokens,
+  chainId: ChainId = getRskChainId(),
+  gasLimit: BigNumberish = MAX_GAS_LIMIT,
+  options?: Partial<CacheCallOptions>,
+): CacheCallResponse<string> => {
+  const gasPrice = useGasPrice(chainId);
+  const balance = useAssetBalance(asset, chainId, null, 0, options);
+
+  const contract = useAsync(() => getTokenContract(asset, chainId));
```
tiltom:
```suggestion
  const contract = useAsync<ContractConfigData>(() => getTokenContract(asset, chainId));
```

tiltom (approved):

soulBit (approved):

PR #219, opened 2023-02-13
SOV-1561: autoconnect to previous wallet

creed-victor:
https://sovryn.atlassian.net/browse/SOV-1561


soulBit (approved):

Rickk137:
@creed-victor  when the app loads or you refresh sometimes the wallet dialog is flashed.

creed-victor:
> @creed-victor when the app loads or you refresh sometimes the wallet dialog is flashed.

issue will be on the onboard package side - something to do with lazy loading most likely not giving up to date state. But I don't think it's an easy and may require great chunk of refactoring.
Looks like it appears only sometimes, so we can ignore it and go back later if it really causes issues

Rickk137 (approved):

PR #238, opened 2023-02-23
SOV-1486: Create AssetValue component

pietro-maximoff:
https://sovryn.atlassian.net/browse/SOV-1486


soulBit (approved):

apps/frontend/src/utils/math.ts
```diff
@@ -73,6 +73,11 @@ export const fromWeiFixed = (
   unitName: BigNumberish = DEFAULT_UNIT,
 ): string => Number(fromWei(value, unitName)).toFixed(decimals);
 
+export const numberFromWei = (
+  amount: any,
+  unitName: BigNumberish = DEFAULT_UNIT,
+) => Number(fromWei(amount, unitName));
+
```
creed-victor: Just use `Number(fromWei(amount, unit));` directly in the AssetValue component.

`Number` only supports 16 decimal places (while tokens have 18) and then rounds up everything so I would prefer not having such helpers exported so devs would not use them by mistake in the transactions.

creed-victor (changes requested):

creed-victor (approved):

PR #250, opened 2023-02-28
SOV-1815: Fix app crashing after disconnecting a wallet on Fast BTC

tiltom:
Resolves https://sovryn.atlassian.net/browse/SOV-1815

### DIFF
--- a/apps/frontend/src/app/3_organisms/FastBtcDialog/FastBtcDialog.tsx
+++ b/apps/frontend/src/app/3_organisms/FastBtcDialog/FastBtcDialog.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useMemo, useState } from 'react';
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
 
 import { t } from 'i18next';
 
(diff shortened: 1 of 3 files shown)


apps/frontend/src/app/3_organisms/FastBtcDialog/FastBtcDialog.tsx
```diff
@@ -79,6 +79,11 @@ export const FastBtcDialog: React.FC<FastBtcDialogProps> = ({
     [isMobile],
   );
 
+  // Reset the selected index to 0 in case the connected account changes because the new account may have the Send tab hidden and it would crash the app if we tried to access that index
```
soulBit, line 82: If possible, can we also close the Funding dialog when last wallet is disconnected? Current behaviour is that dialog remains open in this state (not a big deal since you can't proceed forward, but preferably we close it fully):
image

soulBit (changes requested):
Change works as expected, just one comment to consider

apps/frontend/src/app/3_organisms/FastBtcDialog/FastBtcDialog.tsx
tiltom: Good catch, I just pushed the fix :)

soulBit (approved):

PR #289, opened 2023-03-13
SOV-1987 handle scientific numbers on AmountRenderer

creed-victor:
https://sovryn.atlassian.net/browse/SOV-1987

### DIFF
--- a/apps/frontend/src/app/2_molecules/AmountRenderer/AmountRenderer.tsx
+++ b/apps/frontend/src/app/2_molecules/AmountRenderer/AmountRenderer.tsx
@@ -19,6 +19,7 @@ import {
   formatValue,
   getDecimalPartLength,
   getLocaleSeparators,
+  isScientificNumber,
 } from '../../../utils/math';
 
 const { decimal, thousand } = getLocaleSeparators();
@@ -60,24 +61,29 @@ export const AmountRenderer: FC<AmountRendererProps> = ({
     });
   }, [addNotification, value]);
 
-  const formattedValue = useMemo(() => {
-    const numberValue = Number(value);
-    const decimals = getDecimalPartLength(numberValue);
+  const countUpValues = useMemo(() => {
+    let endValue = String(value);
 
-    if (decimals > precision) {
-      return Number(numberValue.toString().slice(0, -(decimals - precision))); // trimming the difference between the number of decimals we have and the precision we want, cannot use toFixed() as we need to round down
+    if (typeof value === 'number' && isScientificNumber(value)) {
+      endValue = value.toFixed(18);
     }
-    return numberValue;
+
+    const [whole = '', decimals = ''] = endValue.split('.');
+    const end = parseFloat(
+      (whole ?? 0) + '.' + (decimals ?? 0).slice(0, precision),
+    );
+
+    return {
+      end,
+      decimals: getDecimalPartLength(end),
+      thousandSeparator: thousand,
+      decimalSeparator: decimal,
+    };
   }, [precision, value]);
 
   const localeFormattedValue = useMemo(
-    () => formatValue(formattedValue, precision), // we need to format formattedValue to make sure we round down (formatValue function rounds up)
-    [formattedValue, precision],
-  );
-
-  const formattedValueDecimals = useMemo(
-    () => getDecimalPartLength(formattedValue),
-    [formattedValue],
+    () => formatValue(value, precision), // we need to format formattedValue to make sure we round down (formatValue function rounds up)
+    [value, precision],
   );
 
   const valueIsRounded = useMemo(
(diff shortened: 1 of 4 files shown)


apps/frontend/src/app/2_molecules/AmountRenderer/AmountRenderer.tsx
```diff
@@ -60,24 +61,29 @@ export const AmountRenderer: FC<AmountRendererProps> = ({
...
+    let endValue = String(value);
 
-    if (decimals > precision) {
-      return Number(numberValue.toString().slice(0, -(decimals - precision))); // trimming the difference between the number of decimals we have and the precision we want, cannot use toFixed() as we need to round down
+    if (typeof value === 'number' && isScientificNumber(value)) {
+      endValue = value.toFixed(18);
     }
-    return numberValue;
+
+    const [whole = '', decimals = ''] = endValue.split('.');
```
tiltom, line 71:
```suggestion
    const [whole = '', decimals = ''] = endValue.split(decimal);
```
You cannot hardcode a decimal separator like that, it will fail if you pass a localized string with some other decimal separator (in this case, try to pass `10,12` and you'll see). Use `decimal` from `getLocaleSeparators` instead or parse with a regex that accepts both `.` and `,` (but you may have problems with thousands separator).

apps/frontend/src/app/2_molecules/AmountRenderer/AmountRenderer.tsx
```diff
@@ -60,24 +61,29 @@ export const AmountRenderer: FC<AmountRendererProps> = ({
...
+    const end = parseFloat(
+      (whole ?? 0) + '.' + (decimals ?? 0).slice(0, precision),
+    );
+
+    return {
+      end,
+      decimals: getDecimalPartLength(end),
+      thousandSeparator: thousand,
+      decimalSeparator: decimal,
```
tiltom:
```suggestion

```
Unnecessary.

apps/frontend/src/app/2_molecules/AmountRenderer/AmountRenderer.tsx
```diff
@@ -60,24 +61,29 @@ export const AmountRenderer: FC<AmountRendererProps> = ({
...
 
   const localeFormattedValue = useMemo(
-    () => formatValue(formattedValue, precision), // we need to format formattedValue to make sure we round down (formatValue function rounds up)
-    [formattedValue, precision],
-  );
-
-  const formattedValueDecimals = useMemo(
-    () => getDecimalPartLength(formattedValue),
-    [formattedValue],
+    () => formatValue(value, precision), // we need to format formattedValue to make sure we round down (formatValue function rounds up)
```
tiltom:
```suggestion
    () => formatValue(value, precision),
```
The comment is not relevant anymore as you don't format `formattedValue` but it's true that we need to round down. I was looking at `formatValue` and I think there is a bug in it and we should set the default `roundingMode` to `floor`/`trunc` instead of `halfExpand`. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#rounding_modes for the reference.

apps/frontend/src/app/2_molecules/AmountRenderer/AmountRenderer.tsx
```diff
@@ -116,18 +122,16 @@ export const AmountRenderer: FC<AmountRendererProps> = ({
...
-            end={formattedValue}
-            duration={0.7}
-            separator={thousand}
-            decimals={formattedValueDecimals}
-            decimal={decimal}
-            prefix={shouldShowRoundingPrefix ? '~ ' : ''}
-            suffix={` ${suffix.toUpperCase()}`}
-          />
-        </div>
+        <CountUp
```
tiltom: Tooltip doesn't work without the `div` you deleted.

tiltom (changes requested):

apps/frontend/src/app/2_molecules/AmountRenderer/AmountRenderer.tsx
creed-victor: `endValue` is not localised but stringified number (for example using fromWei), so decimal always `.`

tiltom: But if you try to pass e.g. `10,12` to debt in LoC Status, it won't be parsed correctly.

apps/frontend/src/app/2_molecules/AmountRenderer/AmountRenderer.tsx
creed-victor: Got it, but still will wrap both of children to a span instead of div (previous version made animated renderer block element while normal one was inline)

apps/frontend/src/app/2_molecules/AmountRenderer/AmountRenderer.tsx
creed-victor: Checked all usages and I didn't see that there is localised strings passed to the component. And it shouldn't to my opinion. 
To make sure that dev does not pass it by mistake, maybe I could enforce dev to always use `BigNumber` instead with this task: https://sovryn.atlassian.net/browse/SOV-1988

soulBit: @creed-victor good idea, if there's not a million references to update then I say do it (unless tiltom objects)

tiltom: @creed-victor that sounds good

apps/frontend/src/app/2_molecules/AmountRenderer/AmountRenderer.tsx
tiltom: @creed-victor What about this comment, have you looked into it?

creed-victor: @tiltom, changed it to use `trunc`

soulBit (approved):

tiltom (approved):

PR #314, opened 2023-03-22
SOV-1758: Implement new notification signup user flow

tiltom:
Resolves https://sovryn.atlassian.net/browse/SOV-1758

### DIFF
--- a/apps/frontend/src/locales/en/translations.json
+++ b/apps/frontend/src/locales/en/translations.json
@@ -573,12 +573,11 @@
         "errorMessage": "Something went wrong, please try again.",
         "authErrorMessage": "Authentication failed, please try again.",
         "unconfirmedEmailWarning": "Unconfirmed - check inbox for confirmation email",
-        "signatureError": "The signature does not match the user’s connected Rootstock address",
-        "expiredError": "The authentication message signed is not the most recent authentication message that the server provided to the user",
         "invalidEmail": "Invalid email address",
         "unsavedChanges": "Your changes have not been saved yet",
         "savedChanges": "Changes saved",
-        "unsubscribed": "You were successfully unsubscribed"
+        "unsubscribed": "You were successfully unsubscribed",
+        "authenticationFailed": "Authentication failed"
     },
     "termsOfUsePage": {
         "meta": {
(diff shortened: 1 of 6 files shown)


pietro-maximoff (approved):

apps/frontend/src/locales/en/translations.json
```diff
@@ -572,7 +572,11 @@
         "successMessage": "Your preferences have been saved.",
         "errorMessage": "Something went wrong, please try again.",
         "authErrorMessage": "Authentication failed, please try again.",
```
soulBit:
```suggestion
        "authErrorMessage": "Authentication failed",
```

Needs to be updated as per spec - I'm fine with current copy if it's approved by John (but we'll have to update D1 PR to match if so)

apps/frontend/src/locales/en/translations.json
```diff
@@ -572,7 +572,11 @@
...
-        "unconfirmedEmailWarning": "Unconfirmed - check inbox for confirmation email"
+        "unconfirmedEmailWarning": "Unconfirmed - check inbox for confirmation email",
+        "invalidEmail": "Invalid email address",
+        "unsavedChanges": "Your changes have not been saved yet",
+        "savedChanges": "Changes saved",
+        "unsubscribed": "You were successfully unsubscribed"
```
soulBit:
```suggestion
        "unsubscribed": "Unsubscribed"
```

Needs to be updated as per spec - I'm fine with current copy if it's approved by John (but we'll have to update D1 PR to match if so)

soulBit (changes requested):
Two minor things to check with Product

apps/frontend/src/locales/en/translations.json
tiltom: I committed the changes, no need to overcomplicate it and we can always change it in the future.

soulBit (approved):
Looks good tried on test/mainnet and works well, only part that I didn't explicitly test was point 4 from spec - going to confirm with Shadow if that's been implemented as per spec. edit: confirmed that was tested thoroughly on testnet already, can't test on mainnet unless system is in critical state

PR #331, opened 2023-04-04
SOV-1876: disable closing LoC when in Recovery mode

Rickk137:
https://sovryn.atlassian.net/browse/SOV-1876

### DIFF
--- a/apps/frontend/src/app/3_organisms/ZeroLocForm/CloseCreditLine.tsx
+++ b/apps/frontend/src/app/3_organisms/ZeroLocForm/CloseCreditLine.tsx
@@ -149,7 +149,7 @@ export const CloseCreditLine: FC<CloseCreditLineProps> = ({
         />
       </SimpleTable>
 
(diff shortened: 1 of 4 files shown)


apps/frontend/src/app/3_organisms/ZeroLocForm/CloseCreditLine.tsx
```diff
@@ -157,6 +161,14 @@ export const CloseCreditLine: FC<CloseCreditLineProps> = ({
         />
       )}
 
+      {isRecoveryMode && (
```
tiltom, line 163: Adjust the condition on line `153` to `hasError && !isRecoveryMode` because in case both `hasError` and `isRecoveryMode` are true, you will see 2 `ErrorBadge` and the first one is not needed.

tiltom (changes requested):

tiltom (approved):

PR #337, opened 2023-04-11
SOV-1986: fix asset balance refreshing

pietro-maximoff:
https://sovryn.atlassian.net/browse/SOV-1986
in this PR:
 - fix the balance issue 
 - fix chart js issue for mobile

### DIFF
--- a/.changeset/wild-wasps-fix.md
+++ b/.changeset/wild-wasps-fix.md
@@ -2,4 +2,4 @@
 "frontend": patch
 ---
 
-SOV-1986: Weird behavior when the internet connection is not super fast
+SOV-1986: fix asset balance refreshing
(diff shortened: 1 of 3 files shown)


creed-victor (commented):
Issue here was a race condition and previous request completing after props changes?
Looks like current fix working fine, but it may have been adjusted some functionality and other places where balance is retrieved needs to be refreshed (i'm thinking it may have been effected caching and value reusing between components).

I would suggest removing blocking balance retrieval request when it's already loading, and instead check if `hashedArgs` have changed from the one that was used for the request before `setState` is called.

https://github.com/DistributedCollective/sovryn-dapp/pull/337/files#diff-a9d7fd3051c8ac6c9a2096a5a50c9af76e56eaf25f995d1d1a3b63d3d7b406d0R83-R99

tiltom (approved):

creed-victor (approved):

PR #344, opened 2023-04-14
Github workflow to release frontend build to IPFS network

capitalist42:
The main objective of this pull request is to have the frontend accessible in the IPFS network.

CHANGES
- define Github Workflow "Release Frontend build to IPFS"
- include .env.production file for frontend build
- use relative path to resolve static assets
- replace "createBrowserRouter" with "createHashRouter"

The workflow generates a frontend build and pins the content in the IPFS network via Pinata API. And then it creates a Github release with gateway links. (example: https://github.com/capitalist42/sovryn-dapp/releases/tag/d5562d9)
The main reason to use Pinata is that it is easier for anyone to set up the workflow to verify the "ContentID" of the frontend build. 

To make this workflow works, you have add the following secrets in the "Actions secrets and variables" settings
- `PINATA_JWT` 
- `PINATA_KEY` 
- `PINATA_SECRET`

If you have a custom IPFS gateway and want to add it to the GitHub release, you must define `IPFS_GATEWAY_URL` as a variable. 

To load resources easily, a relative path setting is used to resolve the resources. 
Example link to an image in IPFS. 
https://cloudflare-ipfs.com/ipfs/bafybeic6mqgsv5skq5uzlkhbw7rnof2shqljmoem6sriqcsc3w6n43geqm/assets/636103ef1f7e59d03648f98f_Borrow_hero_gradient.jpg

The hash router is used to manage the application URL.
example link to Sovryn Dapp Earn page.
https://cloudflare-ipfs.com/ipfs/bafybeic6mqgsv5skq5uzlkhbw7rnof2shqljmoem6sriqcsc3w6n43geqm/#/earn

If we use a normal URL, it will not display the Earn page. Because there is no file "earn" under the ContentID.  
https://cloudflare-ipfs.com/ipfs/bafybeic6mqgsv5skq5uzlkhbw7rnof2shqljmoem6sriqcsc3w6n43geqm/earn

`REACT_APP_IPFS_BUILD` env variable is introduced to flag IPFS build. When "REACT_APP_IPFS_BUILD" is true, `createHashRouter` will be used instead of `createBrowserRouter`.


soulBit:
Thanks for submitting this @capitalist42! Once it's final, please update the PR description with some details on the changes and mark it as Ready for Review

soulBit (changes requested):
@capitalist42 thanks for submitting this - one last amendment to make before we do full review - please can you run `yarn changeset` from root folder to generate a new changeset file for this PR?

see below for an example of the file this would generate - note the message included here will be immortalized in our changelog, in this case I think you can reuse title of this PR.

https://github.com/DistributedCollective/sovryn-dapp/pull/347/files#diff-42a95a649aacc35cbf41ddf32ac6b1235902e28d8b4ae5589fb10c2af2058bef

apps/frontend/src/router.tsx
```diff
@@ -33,7 +33,7 @@ const RewardsPage = loadable(
   () => import('./app/5_pages/RewardsPage/RewardsPage'),
 );
 
-export const router = createBrowserRouter([
+export const router = createHashRouter([
```
creed-victor: Please add additional environment variable to flag build as an IPFS and only use `createHashRouter` if it exists, otherwise - keep `createBrowserRouter` as a router

creed-victor (changes requested):

tiltom:
Thanks for the PR @capitalist42 :) 

We discussed it within the team and have one suggestion/requirement - could you adjust the GH workflow so it checks if there is some deployment already and delete it before the new one? Basically, we don't want historic versions for 2 reasons:

1. We'll probably just use free plan in the beginning and even if we decide to use some paid one in the future, we don't want to overpay.
2. We don't want to have older versions accessible because there may be bugs.

capitalist42:
The purpose of IPFS is to have historic versions of the web. In our case, we use it to have an immutable version of the "dapp" web interface. The one way to remove or make some content inaccessible is to command all IPFS nodes to remove those content. Once you release a ContentId to the public, some users will pin the ContentId so that the frontend build will be stored in their local IPFS node. 

Right now, the workflow will pin the build content with name metadata via Pinata API. I could add a step to call Pinata API to unpin some content in Github workflow. Do you want the step to unpin all builds? or just a certain build? @tiltom 

I think I should add `frontendVersion` as input to this workflow. And then include the value as metadata when calling API to pin. In this way, a step can be added to unpin all content with the same `frontendVersion`.

.changeset/purple-ravens-compete.md
```diff
@@ -0,0 +1,6 @@
+---
+'frontend': major
```
creed-victor:
```suggestion
'frontend': patch
```

tiltom:
> The purpose of IPFS is to have historic versions of the web. In our case, we use it to have an immutable version of the "dapp" web interface. The one way to remove or make some content inaccessible is to command all IPFS nodes to remove those content. Once you release a ContentId to the public, some users will pin the ContentId so that the frontend build will be stored in their local IPFS node.
> 
> Right now, the workflow will pin the build content with name metadata via Pinata API. I could add a step to call Pinata API to unpin some content in Github workflow. Do you want the step to unpin all builds? or just a certain build? @tiltom
> 
> I think I should add `frontendVersion` as input to this workflow. And then include the value as metadata when calling API to pin. In this way, a step can be added to unpin all content with the same `frontendVersion`.

If users pin a dapp version to their local IPFS nodes, that's fine and there's nothing we can do (they are aware of possible consequences of using older versions). I was talking about unpinning from Pinata as there are limits for the number of pinned files (https://www.pinata.cloud/pricing#plans-details).

It'd be ideal if you could add a step to the GH workflow and unpin all the previous pinned versions. Your proposed solution could work. 

Another way how to do it would be via Pinata API (I haven't tried it) - list all the pinned versions (https://docs.pinata.cloud/pinata-api/pinning/list-pin-by-cid-jobs) and then unpin them (https://docs.pinata.cloud/pinata-api/pinning/remove-files-unpin).

capitalist42:
@tiltom I added an optional step to unpin all previous frontend builds with Pinata API.

soulBit:
@capitalist42 thanks for updating this, we're a bit swamped in current sprint so won't get a chance to finalise review and merge this till late next week at the earliest - apologies if it's taken a little while to get it completed, we haven't forgotten about it :)

creed-victor (approved):

creed-victor:
@capitalist42 , thanks for the PR, please check on the conflicts and we are good for merging this in :)

capitalist42:
@creed-victor I just resolved the merge conflict.

PR #392, opened 2023-05-25
fix: skip approval step if there is enough allowance

creed-victor:


packages/sdk/src/internal/utils.ts
```diff
@@ -27,8 +28,21 @@ export const getMinReturn = (
 const Erc20IFace = new ethers.utils.Interface([
   'function transfer(address to, uint256 amount) public',
   'function approve(address spender, uint256 amount) public',
+  'function allowance(address owner, address spender) public view returns (uint256)',
 ]);
 
+export const testAllowance = async (
```
tiltom:
```suggestion
export const hasEnoughAllowance = async (
```

tiltom (changes requested):

tiltom (approved):

PR #420, opened 2023-06-25
SOV-2436: Stability Pool Page - Wallet Disconnect Data Reset

pietro-maximoff:
https://sovryn.atlassian.net/browse/SOV-2436

### DIFF
--- a/.changeset/six-coins-own.md
+++ b/.changeset/six-coins-own.md
@@ -1,0 +1,5 @@
+---
+"frontend": patch
+---
+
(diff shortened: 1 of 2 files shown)


tiltom (changes requested):
When you disconnect a wallet, you see a warning message flash. It's caused by the fact you don't check if you have a connected wallet on line `438`. Just adjust the condition there to `!isValidAmount && account` and it should be fixed.

tiltom (approved):

Rickk137 (approved):

PR #421, opened 2023-06-25
SOV-2437: Reward Page - Wallet Disconnect Data Reset

pietro-maximoff:
https://sovryn.atlassian.net/browse/SOV-2437

### DIFF
--- a/apps/frontend/src/app/5_pages/RewardsPage/RewardsPage.tsx
+++ b/apps/frontend/src/app/5_pages/RewardsPage/RewardsPage.tsx
@@ -83,16 +83,11 @@ const RewardsPage: FC = () => {
     [amount, claimLocked, signer, sovGain],
   );
 
-  const renderBtcGain = useMemo(
-    () => (account ? amount : Decimal.ZERO),
-    [account, amount],
-  );
-
-  const renderSovGain = useMemo(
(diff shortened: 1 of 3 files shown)


apps/frontend/src/app/5_pages/RewardsPage/RewardsPage.tsx
```diff
@@ -83,6 +83,16 @@ const RewardsPage: FC = () => {
...
 
+  const renderBtcGain = useMemo(
+    () => (account ? amount : Decimal.ZERO),
+    [account, amount],
+  );
+
+  const renderSovGain = useMemo(
+    () => (account ? sovGain : Decimal.ZERO),
+    [account, sovGain],
+  );
```
creed-victor: This only helps if user does not reconnect or reconnects with same wallet that was previously connected, If user would connect with different wallet he would be shown rewards of previous `setAmount` amount until new `liquity.getStabilityDeposit(account)` request is completed.

Correct way on handling it would be to reset `setAmount` to 0 after user disconnects or using `asyncCall` helper (or observeCall + startCall) to make use of value caching for the wallet too

creed-victor (changes requested):

creed-victor (approved):

soulBit (approved):

PR #439, opened 2023-07-09
SOV-2615: new Rewards page total rewards earned

Rickk137:
https://sovryn.atlassian.net/browse/SOV-2615

### DIFF
--- a/.changeset/mighty-mice-fry.md
+++ b/.changeset/mighty-mice-fry.md
@@ -1,0 +1,6 @@
+---
+"frontend": patch
+"@sovryn/ui": patch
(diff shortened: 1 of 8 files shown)


tiltom (changes requested):
Pls fix the following issues:

1. Title has a different color than in the Figma design.
2. The grey background should be full width, regardless of the value.
3. If there's no connected wallet, the value should be `N/A` (it's `0 BTC` currently).
4. There should be more space on top so the tooltip doesn't overlap with the header menu.

tiltom (approved):

PR #445, opened 2023-07-13
SOV-2708: Update Convert page to enable linking to specific pairs

pietro-maximoff:
https://sovryn.atlassian.net/browse/SOV-2708


tiltom (changes requested):
It doesn't work if a token display name is different from the token symbol (e.g. Babelfish - FISH).

https://deploy-preview-445--sovryn-dapp.netlify.app/convert?from=babelfish

pietro-maximoff:
@tiltom yep, right, it's because all tokens have normal enums except fish = "Babelfish", 🤪 
will fix it

apps/frontend/src/app/5_pages/ConvertPage/ConvertPage.tsx
```diff
@@ -78,10 +83,22 @@ const ConvertPage: FC = () => {
...
+    if (
+      fromToken &&
+      fromToken in SupportedTokens &&
+      fromToken !== SupportedTokens.fish
+    ) {
+      return fromToken as SupportedTokens;
+    }
+    if (fromToken === SupportedTokens.fish) {
+      return SupportedTokens.fish;
+    }
```
creed-victor: I think you wanted to rewrite `fish` to `babelfish` here?

```suggestion
    if (fromToken === SupportedTokens.fish) {
      return SupportedTokens.babelfish;
    }
```

creed-victor: if so, rewrite `fish` to `babelfish` for `to` field too, if `fish` comes in as a query param

creed-victor (changes requested):

apps/frontend/src/app/5_pages/ConvertPage/ConvertPage.tsx
pietro-maximoff: @creed-victor but we don't have a `babelfish` in SupportedTokens, it's a value, and key is a `fish`

creed-victor: ah right, then maybe it was supposed to be:
```suggestion
    if (fromToken === "fish") {
      return SupportedTokens.fish;
    }
```
@pietro-maximoff

pietro-maximoff: @creed-victor it won't work that way, because when we select the `fish` token the value will be `babelfish` and this condition `(fromToken === "fish")` won't work

creed-victor (approved):

tiltom (approved):

PR #448, opened 2023-07-13
SOV-2740: lending adjust modal

creed-victor:
https://sovryn.atlassian.net/browse/SOV-2740


creed-victor:
changes already at #479

PR #472, opened 2023-07-29
SOV-2811: update delegate comp

pietro-maximoff:
https://sovryn.atlassian.net/browse/SOV-2811


apps/frontend/src/app/5_pages/StakePage/components/VestingStakesFrame/components/AdjustVestingStakeRenderer.tsx
```diff
@@ -24,6 +26,15 @@ export const AdjustVestingStakeRenderer: FC<Vesting> = ({
...
   );
 
+  const { endDate } = useGetVestingStakeStartEndDates(vestingContract);
+
+  const currentDate = useMemo(() => new Date(), []);
+
+  const isDisabled = useMemo(
+    () => currentDate.getTime() / MS > Number(endDate),
+    [endDate, currentDate],
+  );
```
tiltom:
```suggestion
  const currentDate = useMemo(() => dayjs().unix(), []);

  const isDisabled = useMemo(
    () => currentDate > Number(endDate),
    [endDate, currentDate],
  );
```
You can just use `unix` function from `dayjs` to get a timestamp.

apps/frontend/src/constants/gasLimits.ts
```diff
@@ -16,7 +16,7 @@ export const GAS_LIMIT = {
   STAKING_INCREASE_STAKE: 450_000,
   STAKING_EXTEND: 450_000,
   STAKING_WITHDRAW: 1_400_000,
-  VESTING_DELEGATE: 450_000,
+  VESTING_DELEGATE: 1_690_000,
```
tiltom: I got two failing txs when I tried the default settings, raise the limit to 2000 000 at least. This is my last successful tx and as you can see, it used almost 2000 000 gas.

tiltom (changes requested):

tiltom (approved):

PR #485, opened 2023-08-04
SOV-2801: move fund wallet button to the header

creed-victor:
https://sovryn.atlassian.net/browse/SOV-2801


soulBit (changes requested):
Looks good except for one issue - on deployed link the funding button doesn't use correct styles:
image
(works fine locally)

soulBit (approved):

PR #486, opened 2023-08-05
SOV-2748 / SOV-2974: Develop D2 Borrow page - New loan modal

pietro-maximoff:
https://sovryn.atlassian.net/browse/SOV-2748
https://sovryn.atlassian.net/browse/SOV-2974


apps/frontend/src/app/3_organisms/BorrowLoanForm/components/NewLoanForm/NewLoanForm.tsx
```diff
@@ -0,0 +1,552 @@
...
+  );
+
+  const preparedInterest = calculatePrepaidInterest(
+    borrowApr,
+    borrowSize,
+    borrowDays,
+  );
+
+  const originationFee = useMemo(
+    () => getOriginationFeeAmount(borrowSize, minBorrowingFeeRate),
```
soulBit: According to spec this should be calculated with collateralAmount instead of borrow size. I believe token this is shown as should also be the same as collateral asset.

Thinking about this in more detail, I don't believe we actually have an Origination Fee on the standard Sovryn borrowing (it's not used anywhere else in the PRD, and isn't a param passed to SC calls or returned from loan queries - e.g. in data from https://github.com/DistributedCollective/Sovryn-smart-contracts/tree/development#71-loans), so I think this may be an error in spec - please clarify with Light if this was intended.

apps/frontend/src/app/3_organisms/BorrowLoanForm/components/NewLoanForm/NewLoanForm.tsx
```diff
@@ -0,0 +1,552 @@
...
+
+      <div className="mt-6">
+        <Checkbox
+          checked={hasDisclaimerBeenChecked}
+          onChangeValue={setHasDisclaimerBeenChecked}
+          label={
+            <Trans
+              i18nKey={pageTranslations.newLoanDialog.labels.disclaimer}
+              components={[
+                <Link
```
soulBit: Remove this link if we don't have the URL yet, to avoid leaving it in accidentally later

.changeset/soft-windows-buy.md
```diff
@@ -0,0 +1,5 @@
+---
+"frontend": patch
+---
+
+SOV-2748: Borrow page - New loan modal
```
soulBit, line 5: Since we modify the other packages, I split changeset into multiple files to more accurately reflect what was updated (keeps changelog clear and meaningful)

soulBit (changes requested):

apps/frontend/src/app/3_organisms/BorrowLoanForm/components/NewLoanForm/NewLoanForm.tsx
tiltom: We still don't have the URL, I pinged light on the PRD.

apps/frontend/src/app/3_organisms/BorrowLoanForm/components/NewLoanForm/NewLoanForm.tsx
tiltom: Good catch, I fixed it. I can ask Light about it once he's back but we show `Origination fee` in `Adjust dialog` as well so I guess it's not a mistake.

PR #601, opened 2023-09-22
SOV-3149: Proposal creation modal

Rickk137:
https://sovryn.atlassian.net/browse/SOV-3149

### DIFF
--- a/apps/frontend/src/app/5_pages/BitocracyPage/components/NewProposalForm/NewProposalForm.constants.ts
+++ b/apps/frontend/src/app/5_pages/BitocracyPage/components/NewProposalForm/NewProposalForm.constants.ts
@@ -1,0 +1,82 @@
+import { ProposalType } from './NewProposalForm.types';
+
+export const proposalTypeOptions = [
+  {
+    value: ProposalType.Parameter,
+    label: ProposalType.Parameter,
+  },
+  {
+    value: ProposalType.Proclamation,
+    label: ProposalType.Proclamation,
+  },
+  {
+    value: ProposalType.Treasury,
+    label: ProposalType.Treasury,
+  },
+];
+
+export const proposalContractOptions = [
+  {
+    value: 'Sovryn Protocol',
+    label: 'Sovryn Protocol',
+  },
+  {
+    value: 'Staking',
+    label: 'Staking',
+  },
+  {
+    value: 'LoanTokenLogicLM',
+    label: 'LoanTokenLogicLM',
+  },
+  {
+    value: 'LoanTokenLogicWRBTC',
+    label: 'LoanTokenLogicWRBTC',
+  },
+  {
+    value: 'iDOC',
+    label: 'iDOC',
+  },
+  {
+    value: 'iUSDT',
+    label: 'iUSDT',
+  },
+  {
+    value: 'iRBTC',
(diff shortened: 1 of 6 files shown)


apps/frontend/src/app/5_pages/BitocracyPage/components/NewProposalForm/NewProposalForm.constants.ts
```diff
@@ -0,0 +1,82 @@
+import { ProposalType } from './NewProposalForm.types';
+
+export const proposalTypeOptions = [
```
tiltom:
```suggestion
export const PROPOSAL_TYPE_OPTIONS = [
```
Constants should be uppercased, change all of them in this file.

apps/frontend/src/app/5_pages/BitocracyPage/components/NewProposalForm/NewProposalForm.constants.ts
```diff
@@ -0,0 +1,82 @@
...
+  },
+  {
+    value: ProposalType.Treasury,
+    label: ProposalType.Treasury,
+  },
+];
+
+export const proposalContractOptions = [
+  {
+    value: 'Sovryn Protocol',
```
tiltom: I think this should be an enum as well because you'll want to compare it somewhere and you don't want to have magic strings.

apps/frontend/src/app/5_pages/BitocracyPage/components/NewProposalForm/NewProposalForm.constants.ts
```diff
@@ -0,0 +1,82 @@
...
+  },
+  {
+    value: 'iXUSD',
+    label: 'iXUSD',
+  },
+];
+
+export const proposalTreasuryOptions = [
+  {
+    value: 'Sovryn Protocol',
```
tiltom: I think this should be an enum as well because you'll want to compare it somewhere and you don't want to have magic strings.

apps/frontend/src/app/5_pages/BitocracyPage/components/NewProposalForm/NewProposalForm.tsx
```diff
@@ -0,0 +1,56 @@
...
+import { translations } from '../../../../../locales/i18n';
+import {
+  proposalContractOptions,
+  proposalTreasuryOptions,
+  proposalTypeOptions,
+} from './NewProposalForm.constants';
+import { ProposalType } from './NewProposalForm.types';
+
+type NewLoanFormProps = {
+  onSuccess: () => void;
```
tiltom: You don't use this prop at all. Also, I am not sure if this should be `onSuccess` or `onSubmit`, what do you think? Maybe we can provide both and have them optional?

tiltom (changes requested):

apps/frontend/src/app/5_pages/BitocracyPage/components/NewProposalForm/NewProposalForm.tsx
Rickk137: This is something we can add when when needed. so i'll remove it for now

tiltom (approved):

PR #603, opened 2023-09-22
SOV-3196: add voting tooltips to Voting proposals page

Rickk137:
https://sovryn.atlassian.net/browse/SOV-3196


tiltom (changes requested):
The code looks fine but I'd add some margin between the vote start date and vote end label.

image

PR #623, opened 2023-10-04
SOV-3236: proposal creation state

creed-victor:
https://sovryn.atlassian.net/browse/SOV-3236


soulBit (approved):

apps/frontend/src/app/5_pages/BitocracyPage/components/NewProposalForm/NewProposalForm.tsx
```diff
@@ -46,6 +82,7 @@ export const NewProposalForm: FC = () => {
         <Button
           text={t(translations.common.buttons.continue)}
           className="w-full sm:w-auto"
+          onClick={() => submit()}
```
Rickk137, line 85: onClick={submit}

Rickk137 (changes requested):

apps/frontend/src/app/5_pages/BitocracyPage/components/NewProposalForm/NewProposalForm.tsx
creed-victor: submit is not intended to use directly by onClick and args are not valid for button event so made like this for now. Added here just like an example and actual submit handler should be implemented in actual modal.

PR #628, opened 2023-10-05
SOV-3223: Close menu when menu option is clicked

pietro-maximoff:
https://sovryn.atlassian.net/browse/SOV-3223


soulBit (changes requested):
This only closes the individual dropdown e.g. Earn (when Lend is clicked) - we should close entire mobile menu when an option is selected

soulBit (approved):

PR #801, opened 2024-03-07
SOV-3777: boltz v2 api

creed-victor:
https://sovryn.atlassian.net/browse/SOV-3777


apps/frontend/src/app/3_organisms/FastBtcDialog/utils/boltz/boltz.client.ts
```diff
@@ -0,0 +1,164 @@
...
+  SubmarineSwapResponse,
+  SubmarineSwapTransactionResponse,
+  SwapResponse,
+} from './boltz.types';
+
+const baseUrl = isMainnet()
+  ? 'https://api.boltz.exchange/api/v2'
+  : 'https://testnet.boltz.exchange/api/v2';
+
+const client = Axios.create({
```
rick23p, line 28: boltzClient? and doesn't need to be exported

apps/frontend/src/app/3_organisms/FastBtcDialog/utils/boltz/boltz.types.ts
```diff
@@ -0,0 +1,183 @@
+export enum StatusEnum {
```
rick23p: I think this can be better named

rick23p (changes requested):

apps/frontend/src/app/3_organisms/FastBtcDialog/utils/boltz/boltz.client.ts
creed-victor: It's being wrapped by functions in this file and does not used for outside it, so i don't think we need need it exported

apps/frontend/src/app/3_organisms/FastBtcDialog/utils/boltz/boltz.types.ts
creed-victor: Open to suggestions, just that StatusType are already used in the fastbtc flows

rick23p: Maybe `BoltzTxStatus`

rick23p (approved):

PR #835, opened 2024-04-11
SOV-4014: claim LP page

creed-victor:
https://sovryn.atlassian.net/browse/SOV-4014


soulBit (approved):
Still needs mainnet details added, we'll do that on bob-base

PR #1010, opened 2024-09-20
money market qa fixes

chescalante:
**APY vs. APR**

> We wanted to provide clarity regarding the APY vs. APR values in the mobile version of the Borrow Assets table. Currently, all displayed values are APY, and we have updated the label to correctly reflect this. Additionally, we removed the value from the expanded accordion view, as switching to APR here, as indicated in the designs, would introduce inconsistency.
> 
> On the desktop version, APY is displayed throughout, including in the title. Showing APR only within the expanded accordion on mobile may not align well with the rest of the interface.
> 
> Please let us know if you'd prefer to proceed with the APR change on mobile, or if the current setup works for your needs.

**View Transactions Button**

> Although the 'View Transactions' button is included in the designs, no specific functionality has been defined for it. As a result, we decided not to include it at this stage.


apps/frontend/src/app/2_molecules/AavePoolRowTitle/AavePoolRowTitle.tsx
```diff
@@ -9,13 +9,15 @@ import {
...
 type AaveRowTitleProps = AmountRendererProps & {
+  isOpen?: boolean;
   asset: string;
   label?: string;
 };
 
 export const AaveRowTitle: FC<AaveRowTitleProps> = ({
   asset,
   label,
+  isOpen,
```
pietro-maximoff: you can define `isOpen = false` by default and then you don't need to set `isOpen || false` each time you call the AaveRowTitle component

apps/frontend/src/utils/aave/AaveUserReservesSummary.ts
```diff
@@ -209,9 +209,10 @@ export class AaveUserReservesSummaryFactory {
           );
 
           // how much the user can borrow if there's no limit of supply
-          const canBorrow = borrowPower
+          let canBorrow = borrowPower
             .sub(borrowBalance)
             .div(r.reserve.priceInUSD);
+          canBorrow = canBorrow.gt(0) ? canBorrow : Decimal.ZERO;
```
pietro-maximoff:
```suggestion
          const canBorrow = borrowPower
            .sub(borrowBalance)
            .div(r.reserve.priceInUSD)
            .max(Decimal.ZERO);
```
you can use `max` to directly handle the condition of whether the result is greater than zero

apps/frontend/src/hooks/aave/useAaveReservesHistory.tsx
```diff
@@ -0,0 +1,152 @@
...
+  return requestPromise;
+};
+
+const resolutionForTimeRange = (
+  timeRange: ReserveRateTimeRange,
+): RatesHistoryParams => {
+  switch (timeRange) {
+    case ESupportedTimeRanges.OneMonth:
+      return {
+        from: dayjs().subtract(30, 'day').unix(),
```
pietro-maximoff: maybe makes sense to update `dayjs().subtract(30, 'day')` to `dayjs().subtract(1, 'month')` to handle full months for `OneMonth`

pietro-maximoff (changes requested):

PR #1015, opened 2024-09-30
SOV-4468: Show protocol data in BTC and USD only

rick23p:
https://sovryn.atlassian.net/browse/SOV-4468

### DIFF
--- a/apps/frontend/src/hooks/useGetNativeTokenPrice.ts
+++ b/apps/frontend/src/hooks/useGetNativeTokenPrice.ts
@@ -2,6 +2,8 @@ import { useMemo } from 'react';
 
 import axios from 'axios';
 
+import { ChainId } from '@sovryn/ethers-provider';
+
 import { RSK_CHAIN_ID } from '../config/chains';
 
 import { BITCOIN, ETH } from '../constants/currencies';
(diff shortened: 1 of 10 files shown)


apps/frontend/src/hooks/useGetNativeTokenPrice.ts
```diff
@@ -11,24 +13,24 @@ import { useCacheCall } from './useCacheCall';
 import { useCurrentChain } from './useChainStore';
 import { useDollarValue } from './useDollarValue';
 
-export const useGetNativeTokenPrice = () => {
+export const useGetNativeTokenPrice = (token?: 'BTC' | 'ETH') => {
```
tiltom: I wouldn't pass `token` as an optional param here but `chainId`. It will be more readable.

apps/frontend/src/hooks/useGetNativeTokenPrice.ts
```diff
@@ -11,24 +13,24 @@ import { useCacheCall } from './useCacheCall';
...
   const currentChainId = useCurrentChain();
 
   const { value: rBTCPrice } = useCacheCall(
```
tiltom, line 19:
```suggestion
  const { value: nativeTokenPrice } = useCacheCall(
```

tiltom (changes requested):

apps/frontend/src/hooks/useGetNativeTokenPrice.ts
```diff
@@ -2,33 +2,34 @@ import { useMemo } from 'react';
...
 import { getIndexerUrl } from '../utils/helpers';
 import { toWei } from '../utils/math';
 import { useCacheCall } from './useCacheCall';
-import { useCurrentChain } from './useChainStore';
 import { useDollarValue } from './useDollarValue';
 
-export const useGetNativeTokenPrice = () => {
-  const currentChainId = useCurrentChain();
-
+export const useGetNativeTokenPrice = (chainId: ChainId) => {
```
tiltom:
```suggestion
export const useGetNativeTokenPrice = (chainId?: ChainId) => {
```
This should be only optional and you can use either `chainId` if provided or `currentChainId` if not. Otherwise, you're forced to change all the occurrences and you usually don't want to override `currentChainId`.

tiltom (changes requested):

tiltom (approved):

PR #1017, opened 2024-10-03
QA Fixes (second revision)

chescalante:
Fixes included:

- A small amount of ETH on the user’s balance causes all transactions to fail.
- APR is displayed in the Borrow dialog instead of APY.
- There is a slight difference in the APY percentage displayed in the “Your Loans”.
- When fully repaying a loan, a very small residual amount (“dust”) is left, even though the user repaid the full loan amount.
- The transaction fails when the user attempts to withdraw the maximum amount from the lending pool.
- The transaction fails when the user attempts to borrow the maximum amount.
- Fix typo adding deposits/loans text


apps/frontend/src/app/5_pages/AavePage/AavePage.utils.tsx
```diff
@@ -157,3 +158,6 @@ export const sortRowsByOrderOptions = (
       : -direction,
   );
 };
+
+export const formatFloorNumber = (num: Decimal, precision = 2) =>
+  Math.floor(num.toNumber() * 10 ** precision) / 10 ** precision;
```
pietro-maximoff: We already have similar functions, such as `formatValue` or others, in the `apps/frontend/src/utils/math.ts` file. Please double-check and use them instead of creating new ones.

pietro-maximoff (changes requested):

PR #1044, opened 2024-11-04
SOV-4537: Add incentives column to market making

tiltom:
Resolves https://sovryn.atlassian.net/browse/SOV-4537

### DIFF
--- a/apps/frontend/src/app/5_pages/MarketMakingPage/components/AmbientMarketMaking/components/AmbientPoolsTable/components/Incentives/components/Spice/Spice.tsx
+++ b/apps/frontend/src/app/5_pages/MarketMakingPage/components/AmbientMarketMaking/components/AmbientPoolsTable/components/Incentives/components/Spice/Spice.tsx
@@ -1,0 +1,10 @@
+import React, { FC } from 'react';
+
+import bobLogo from '../../../../../../../../../../../assets/chains/bob-light.svg';
+
+export const Spice: FC = () => (
+  <div className="flex items-center ml-2">
+    <img src={bobLogo} alt="BOB" className="w-3.5" />
+    <div className="ml-1.5">Spice</div>
+  </div>
+);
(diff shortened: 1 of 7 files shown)


apps/frontend/src/app/5_pages/MarketMakingPage/components/AmbientMarketMaking/components/AmbientPoolsTable/components/Incentives/components/Spice/Spice.tsx
```diff
@@ -0,0 +1,10 @@
+import React, { FC } from 'react';
+
+import bobLogo from '../../../../../../../../../../../assets/chains/bob-light.svg';
+
+export const Spice: FC = () => (
+  <div className="flex items-center ml-2">
+    <img src={bobLogo} alt="BOB" className="w-3.5" />
+    <div className="ml-1.5">Spice</div>
```
pietro-maximoff, line 8: I believe we should move `Spice` to trans file

pietro-maximoff (changes requested):

pietro-maximoff:
at some point of the screen, you won't be able to click on the Deposit button, not critical, but we should think about how to prevent this, maybe change the table @media by switching from lg to xl or smth else
screenshot-2024-11-06-15 28 53

apps/frontend/src/app/5_pages/MarketMakingPage/components/AmbientMarketMaking/components/AmbientPoolsTable/components/Incentives/components/Spice/Spice.tsx
tiltom: I don't think it's necessary because it's BOB terminology so I don't think they would use something different for Spanish for instance.

pietro-maximoff (approved):

